PR #3, opened 2026-01-08
chore: update libcoraza api

fzipi:
## what

- update to match the latest libcoraza api

## why

- test if it builds and works

### DIFF
--- a/t/coraza-config-merge.t
+++ b/t/coraza-config-merge.t
@@ -11,6 +11,7 @@ use strict;
 
 use Test::More;
 use Socket qw/ CRLF /;
+use IO::Socket::INET;
 
 BEGIN { use FindBin; chdir($FindBin::Bin); }
 
--- a/t/coraza-proxy-h2.t
+++ b/t/coraza-proxy-h2.t
@@ -10,6 +10,7 @@ use warnings;
 use strict;
 
 use Test::More;
+use IO::Socket::INET;
 
 BEGIN { use FindBin; chdir($FindBin::Bin); }
 
--- a/t/coraza-proxy.t
+++ b/t/coraza-proxy.t
@@ -9,6 +9,7 @@ use warnings;
 use strict;
 
 use Test::More;
+use IO::Socket::INET;
 
 BEGIN { use FindBin; chdir($FindBin::Bin); }
 
--- a/t/coraza-request-body.t
+++ b/t/coraza-request-body.t
@@ -11,6 +11,7 @@ use strict;
 
 use Test::More;
 use Socket qw/ CRLF /;
+use IO::Socket::INET;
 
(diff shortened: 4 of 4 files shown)


ppomes:
## Bug: `ngx_str_to_char()` passes pointer by value

While testing this PR, I found that `ngx_str_to_char()` in `src/ngx_http_coraza_utils.c` has a bug: the `str` parameter is passed by value (`char *str`), so the caller never receives the allocated string. Every string passed to libcoraza (rules, URI, method, client/server addresses, file paths) ends up being NULL.

### The fix

**`src/ngx_http_coraza_utils.c`** — change signature to `char **str`, remove the bogus `free()` on pool-allocated memory:

```c
ngx_int_t
ngx_str_to_char(ngx_str_t a, char **str, ngx_pool_t *p)
{
    if (a.len == 0)
    {
        *str = NULL;
        return NGX_OK;
    }

    *str = ngx_pnalloc(p, a.len + 1);
    if (*str == NULL)
    {
        dd("failed to allocate memory to convert ngx_string to C string");
        return NGX_ERROR;
    }
    ngx_memcpy(*str, a.data, a.len);
    (*str)[a.len] = '\0';

    return NGX_OK;
}
```

**`src/ngx_http_coraza_common.h`** — update declaration:

```c
ngx_int_t ngx_str_to_char(ngx_str_t a, char **str, ngx_pool_t *p);
```

**All 7 call sites** need `&` added (in `ngx_http_coraza_module.c`, `ngx_http_coraza_rewrite.c`, `ngx_http_coraza_pre_access.c`):

```diff
-    if (ngx_str_to_char(value[1], rules, cf->pool) != NGX_OK) {
+    if (ngx_str_to_char(value[1], &rules, cf->pool) != NGX_OK) {
```

### Test results

I built the Docker image with this PR + the fix above. Comparison of coraza test results:

| Test | PR #3 only | PR #3 + fix | 
|------|-----------|-------------|
| `coraza-proxy-h2.t` | 12/25 failed | 1/25 failed |
| `coraza-h2.t` | 13/22 failed | 5/22 failed |
| `coraza-proxy.t` | 16/25 failed | 8/25 failed |
| `coraza.t` | 17/23 failed | 12/23 failed |
| `coraza-config.t` | 9/11 failed | 6/11 failed |

The proxy/h2 tests improve dramatically. The remaining failures are related to redirect handling (removed in this PR), audit log config validation (`secauditlogparts` requires trailing Z), and request body intervention timing (`header already sent` alerts).

### Dockerfile note

`prove . -t coraza*.t` runs the entire nginx test suite (456 files) plus coraza tests, because `.` matches everything. Not sure if this is intentional (validating the module doesn't break standard nginx) or a bug. If intentional, it will always fail due to unrelated test failures in the docker environment.

PR #44, opened 2026-04-21
ci: add full CRS regression tests via go-ftw

ppomes:
## Summary
- Run the full CRS 4.25 regression test suite (excluding 920 protocol enforcement) against nginx+Coraza using go-ftw in cloud mode
- Reuses the existing CRS download and nginx setup from CI
- 920 tests excluded because some malformed HTTP requests crash the go-ftw connection through nginx

## Test plan
- CI passes

cc @jcchavezs

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -196,9 +196,12 @@ jobs:
             sleep 1
           done
 
-          # Run go-ftw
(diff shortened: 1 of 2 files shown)


ppomes:
cc @fzipi

fzipi (approved):

Issue #53, opened 2026-06-24
WebSocket upgrade fails when the Coraza module is enabled (`coraza on`)


janfuhrer:
Thanks for maintaining this connector.

We use `coraza-nginx` (via the `ghcr.io/coreruleset/coraza-crs:nginx` image) in front of services that rely on WebSockets. With the module enabled (`coraza on;`), WebSocket upgrades fail. The exact same nginx config with `coraza off;` works. This happens even with the engine in `DetectionOnly`, so it does not appear to be a blocking rule; the connection simply never upgrades.

This looks analogous to the Caddy connector issue https://github.com/corazawaf/coraza-caddy/issues/78, fix in progress in https://github.com/corazawaf/coraza-caddy/pull/262.

## Environment

- `coraza-nginx` via `ghcr.io/coreruleset/coraza-crs:nginx` (bundles this connector + CRS).
- Standard nginx WebSocket proxying (`proxy_http_version 1.1`, `Upgrade` / `Connection` headers).

## Reproduction

`compose.yaml`:
```yaml
services:
  echo:
    image: jmalloc/echo-server:latest   # WebSocket echo at /.ws
  waf:
    image: ghcr.io/coreruleset/coraza-crs:nginx
    depends_on: [echo]
    environment:
      CORAZA_RULE_ENGINE: "DetectionOnly"
    ports: ["8080:8080"]
    volumes: ["./nginx.conf:/config/nginx/nginx.conf:ro"]
```

`nginx.conf`:
```nginx
load_module "/usr/lib/nginx/modules/ngx_http_coraza_module.so";
worker_processes auto;
pid /tmp/nginx.pid;
error_log /dev/stderr notice;
events {}
http {
  coraza on;                                                # control: change to `coraza off;`
  coraza_rules_file /opt/coraza/config/coraza-rules.conf;
  map $http_upgrade $connection_upgrade { default upgrade; '' close; }
  upstream backend { server echo:8080; }
  server {
    listen 8080 default_server;
    location / {
      proxy_pass http://backend;
      proxy_http_version 1.1;
      proxy_set_header Upgrade    $http_upgrade;
      proxy_set_header Connection $connection_upgrade;
      proxy_set_header Host       $host;
    }
  }
}
```

Bring it up, the proxy is then reachable at http://localhost:8080 (backend behind the WAF):
```bash
docker compose up -d
```

**Test**

- Open http://localhost:8080/.ws in your browser and check the web developer tools
[...]

PR #54, opened 2026-06-24
fix: pass through 101 Switching Protocols (WebSocket upgrades)

ppomes:
Related to #53.

With `coraza on`, WebSocket upgrades hang and time out — even in `DetectionOnly`. The header filter delays response headers until phase-4 body inspection completes, but a `101 Switching Protocols` response has no HTTP body (the connection becomes a raw tunnel), so the flush never fires and the handshake is held forever.

Fix: skip the delay for `101`, same as `header_only`/`error_page`.

Reproduced locally (nginx 1.28 + libcoraza 1.6.0, proxying to a backend that returns 101):
- before: WebSocket upgrade → `TimeoutError`
- after: `101 Switching Protocols`, tunnel echoes data through
- SQLi still returns 403 and clean requests 200 with engine `On` — inspection unaffected

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of `101 Switching Protocols` so protocol upgrades (e.g., WebSocket) complete reliably without header delay.
  * Prevented upgrade/tunneling flows from being interrupted by response-header timing changes.
* **Tests**
  * Added a WebSocket-focused regression test to verify upgrade responses return `101 Switching Protocols` promptly and normal HTTP requests still return `200 OK`.

### DIFF
--- a/src/ngx_http_coraza_header_filter.c
+++ b/src/ngx_http_coraza_header_filter.c
@@ -505,8 +505,15 @@ ngx_http_coraza_header_filter(ngx_http_request_t *r)
      * We also skip the delay when body inspection is not needed
      * (SecResponseBodyAccess Off or Content-Type mismatch): in that case
      * there is no phase-4 buffering and the response must not be held back.
+     *
+     * We also skip 101 Switching Protocols: an upgraded connection (e.g.
+     * WebSocket) becomes a raw bidirectional tunnel with no HTTP response
+     * body, so the body filter never sees a last_buf to trigger the flush.
+     * Delaying the 101 would hold the handshake forever and the upgrade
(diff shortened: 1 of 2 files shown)


ppomes:
@fzipi, need your review please ;-)

fzipi (dismissed):
Makes sense. Not for now, but maybe we can add an e2e test for this?

ppomes:
@fzipi - Added a regression test (`t/coraza-websocket.t`) — a Perl backend returns a real 101 on upgrade, the test asserts the handshake completes. Confirmed it fails against the old code (101 held back → timeout) and passes with the fix.

fzipi (approved):

PR #57, opened 2026-06-28
fix: skip deleted (hash == 0) headers when collecting them for coraza

ppomes:
nginx can leave invalidated entries (`hash == 0`) in its request/response header lists. The connector walks the whole list and forwards every entry's key/value pointers to coraza, including the deleted ones whose pointers may be stale.

nginx skips `hash == 0` entries in its own header filter; this does the same in both the request loop (`rewrite`) and the response loop (`header_filter`).

Defensive hardening — not tied to a specific reported crash.


fzipi (approved):
This is too internal to actually make tests for it, probably.

Issue #46, opened 2026-06-01
nginx return in location section causes the connection to close without data.


aunxx:
Hi All,
I've found something odd in testing the v0.11.0 version of the coraza-nginx module.
I am also using v1.6.0 of libcoraza.

All 200 responses seem to work correctly through the WAF, but the following sections within the nginx.conf config file cause the response to be closed without any data.
`
		location = / {
			return 301 /wiki/Main_Page;
		}
`
I have tested various configs, and I can see a similar closed response when using a few return values -- I've tested 301 and 404 and both fail rather than providing the expected return code.

When I reverted to the previous coraza-nginx tag (v0.10.1) the 301 responses are working as expected. 
Bisecting the git commits gives:
b6a8e04e013d30435e5a8f4fffbab10ef014120d -- Working
c1182015e422f22ab0dbb4b49aa353e91e287ae3 -- Fails
which seems to be relating to changes around body responses if not required. 
I wonder if (as there isn't a body in the return from nginx) a zero length body, or undefined body is causing this issue.

I also see the following in the nginx error logs:
`
2026/06/01 00:20:43 [alert] 101386#101386: *1 zero size buf in writer t:0 r:0 f:0 0000000000000000 000055EA9C3DC5B4-000055EA9C3DC5B4 0000000000000000 0-0,
`
which are not present in the previous working version.
audit logs do show the expected values:
`--IXhtTshJje-F--
HTTP/1.1 301 Moved Permanently
date: Sun, 31 May 2026 22:34:36 GMT
connection: close
location: /wiki/Main_Page`

Using an openssl s_client test I see:
`40D7D61F477F0000:error:0A000126:SSL routines::unexpected eof while reading:ssl/record/rec_layer_s3.c:698:
`
as a response.

aunxx:
Hi All,
Just an update:
If I alter the location code to the following:
`                location = / {
                        return 301 /wiki/Main_Page;
                        coraza off;
                }
`
the response is as expected, and returns the 301 to the request.

fzipi:
CC: @ppomes

ppomes:
Hello @aunxx, thanks for the detailed bisect.

I tried to reproduce this locally with a minimal config (just `coraza on` + `SecRuleEngine On` + a `location = / { return 301 ...; }`) on the v0.11.0 module, and the 301 was forwarded correctly — full 169-byte body, no `zero size buf in writer` warning. Same result with `SecResponseBodyAccess Off`.

[...]

Issue #41, opened 2026-04-09
Responses over ~60KB hang/timeout even with SecResponseBodyAccess Off


qfiber:
## Description

The coraza-nginx module causes nginx worker threads to hang on any HTTP response larger than approximately 60KB, even when `SecResponseBodyAccess Off` is explicitly configured. This makes the module unusable for any site serving typical web assets (JS bundles, CSS, WASM, images) or dynamically generated pages.

## Environment

- **OpenResty**: 1.29.2.3
- **libcoraza**: v1.3.0 (commit 627e13a)
- **coraza-nginx**: latest main (commit a1efb80)
- **CRS**: OWASP CRS 4.25.0
- **OS**: AlmaLinux 9 (Linux 6.19.x, x86_64)

## Minimal reproduction

```nginx
# nginx.conf
load_module /usr/local/openresty/nginx/modules/coraza-nginx.so;

http {
    coraza on;
    coraza_rules_file /etc/openresty/coraza/coraza.conf;

    server {
        listen 443 ssl;
        server_name example.com;
        root /var/www/html;
        location / { try_files $uri $uri/ =404; }
    }
}
```

```
# coraza.conf — minimal, everything disabled
SecRuleEngine On
SecRequestBodyAccess Off
SecResponseBodyAccess Off
SecAuditEngine Off
```

```bash
# Create test files of various sizes
dd if=/dev/urandom of=/var/www/html/small.js bs=1K count=50
dd if=/dev/urandom of=/var/www/html/large.js bs=1K count=100

# 50KB works fine
curl -o /dev/null -w "%{http_code}" --max-time 5 https://example.com/small.js
# → 200 (instant)

# 100KB hangs until timeout
curl -o /dev/null -w "%{http_code}" --max-time 10 https://example.com/large.js
# → 000 (timeout, no response)
```

## Findings

| File size | CRS loaded | SecResponseBodyAccess | SecRequestBodyAccess | SecAuditEngine | Result |
|-----------|-----------|----------------------|---------------------|---------------|--------|
| 50 KB | Yes | Off | On | RelevantOnly | **200** (OK) |
| 60 KB | Yes | Off | On | RelevantOnly | **200** (OK) |
| 80 KB | Yes | Off | On | RelevantOnly | **Timeout** |
| 100 KB | Yes | Off | On | RelevantOnly | **Timeout** |
| 100 KB | **No** (no rules) | Off | **Off** | **Off** | **Timeout** |
[...]

PR #61, opened 2026-07-06
ci: add dynamic-analysis workflows (fuzz, sanitizers, valgrind) and regression tests

eilandert:

**Severity:** Low

## What
Two-tier CI plus regression tests for previously fixed upstream issues:

- **ci-fast**: libFuzzer run (value-profile, dictionary, seed corpus), an ASan/UBSan soak of the module under nginx, and an `nginx -t` config lint of the shipped `coraza.conf`.
- **ci-deep** (scheduled/dispatch): valgrind memcheck + helgrind soaks and a longer fuzz run across a server/arch matrix (nginx + angie, amd64 + arm64), with source/Go/ccache caching.
- Static scanners (flawfinder, clang-tidy with a real `compile_commands.json`, semgrep) now gate the build instead of having their exit codes ignored.

## Why
The connector crosses a C↔cgo boundary on every request; sanitizer/valgrind soaks and fuzzing are the only way to catch memory and threading regressions before they ship. Scanner exit codes were previously swallowed, so findings never failed CI.

## Testing
Adds `t/` regression tests covering upstream issues #40, #41, #42, #46 and #54, and fixes two pre-existing `t/` bugs (an `absolute_redirect` URL match; a missing `SecResponseBodyMimeType` that made the response-body tests pass vacuously).

## Summary by CodeRabbit

* **New Features**
  * Added **CI Fast** and **CI Deep** workflows with fuzzing, sanitizer/valgrind analysis, and static/security scanning.
  * Introduced a **soak-test harness** for concurrent traffic against dynamic connector builds.
  * Added libFuzzer tooling plus an expanded fuzz dictionary and corpus.

* **Bug Fixes**
  * Added regression tests for redirects, **101 Switching Protocols**, **Unix socket** mapping, and **response-body access off** (including large-response stability).

* **Documentation**
  * Documented fuzzing setup and how to run it.

* **Chores**
  * Improved CI speed with persistent caching and updated build steps; added Valgrind suppressions and updated local ignore rules.

---

## Update: additional coverage + a debug-build gate

A follow-up audit of the closed PRs against the source turned up test gaps this branch is the right home for. Added:

**`t/coraza-request-body-chunked.t`** — chunked transfer-encoding had **no** coverage anywhere in the suite (every other body test sends a `Content-Length`), and the temp-file branch in `pre_access.c` — which hands Coraza a file path via `coraza_request_body_from_file()` instead of appending buffers — was never entered, because every existing body test sends a few hundred bytes and stays in memory. Covers both, including a marker straddling the `client_body_buffer_size` boundary.

Verified as a real guard by mutation rather than assumed: stubbing out `coraza_request_body_from_file()` fails exactly the three temp-file assertions and leaves the controls green.

**`t/coraza-deleted-headers.t`** — nginx marks a deleted header by zeroing its hash and may leave key/value pointers stale. The response-side skip is reachable via `proxy_hide_header` and is asserted both ways: the hidden header is invisible to the WAF, plus a control proving the same rule *does* fire when it is not hidden.

The request-side skip is **not** reachable from configuration alone (`proxy_set_header` rewrites the upstream request, it does not touch `headers_in`), so that half asserts full list traversal instead. The file says so in a comment rather than implying coverage it does not have.

**`t/coraza-empty-header-value.t`** — empty header values and empty query strings. Scoped honestly in the header comment: header values do **not** flow through `ngx_str_to_char()` (they go to `coraza_add_request_header` as explicit `(ptr, len)` pairs), so these pass both before and after the zero-length fix and are not a regression guard for it.

**`ci-fast: debug-build`** — the `dd()` call sites are compiled out of every other job, so a broken one ships unnoticed: an undeclared variable inside `dd()` once made every `--with-debug` build fail to compile with no CI job to catch it. This job compiles with `--with-debug` and `-DCORAZA_DDEBUG=1` under `-Werror`. Verified it fails on a reintroduced bad `dd()` call and passes on the current tree.

Full suite on this branch: 28 files, 323 assertions, green.

Note on the five source-grep tests (`coraza-config-flags.t`, `coraza-dynlib-contract.t`, `coraza-forward-header-rc.t`, `coraza-header-error-propagation.t`, `coraza-delayed-file-buffer.t`): they read `../config` and `../src/*`, so they only pass when `prove` runs from inside a checkout. Copied into the `nginx-tests` tree the way `build.yml` does it, they abort. Pre-existing, not introduced here, but worth knowing before reading a CI log.


eilandert:
Addressed the CodeRabbit findings in f3389a6:

- `persist-credentials: false` on all 7 checkout steps (ci-deep, ci-fast).
- ci-deep: `workflow_dispatch` fuzz inputs now routed through `env` instead of inline `${{ }}` interpolation.
- ci-fast: `CORAZA_MODULE_SO` now derives from `env.NGINX_VERSION` (was hardcoded `nginx-1.28.0`).
- ci-fast: config-lint comment reworded to match actual coverage (shipped `coraza.conf` only).
- ci-deep: replaced the `ls objs/nginx objs/angie` candidate pick with an explicit existence loop (SC2012).
- `fuzz/run.sh`: `cd "$FUZZ_DIR" || exit 1`.
- `tools/soak.sh`: EXIT trap now kills the (possibly valgrind-wrapped) server before removing the workdir, so an early failure can't orphan it.
- package.yml: dropped the no-op `Configure environment` export step (consistent with the same removal in build.yml).

Skipped: checksum pinning of source downloads (pinned tags over HTTPS; can do as a follow-up if wanted) and the pre-existing shellcheck warnings in build.yml's CRS/go-ftw step, which this PR doesn't touch.

ppomes:
@coderabbitai review

.github/workflows/ci-fast.yml
```diff
@@ -0,0 +1,283 @@
...
+#   sanitizer-lite  : 60s ASan/UBSan soak of the running module under benign +
+#                     attack traffic (valgrind/helgrind are deep-only).
+# Long discovery (parallel multi-hour fuzz, full memcheck+helgrind soak,
+# scanners) lives in ci-deep.yml (monthly cron + workflow_dispatch).
+
+env:
+  FORCE_JAVASCRIPT_ACTIONS_TO_NODE24: true
+  # renovate: depName=libcoraza packageName=corazawaf/libcoraza datasource=github-releases
+  LIBCORAZA_VERSION: v1.4.0
+  NGINX_VERSION: "1.28.0"
```
fzipi: Can we use latest stable? 1.30.x?

.github/workflows/ci-fast.yml
```diff
@@ -0,0 +1,283 @@
...
+  group: ci-fast-${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+
+jobs:
+  fuzz-regression:
+    name: Fuzz regression (120s)
+    runs-on: ubuntu-22.04
```
fzipi: Can we use 24.04?

.github/workflows/ci-fast.yml
```diff
@@ -0,0 +1,283 @@
...
+        if: failure()
+        uses: actions/upload-artifact@330a01c490aca151604b8cf639adc76d48f6c5d4 # v5
+        with:
+          name: fuzz-fast-crashes
+          path: ${{ github.workspace }}/fuzz/crash-*
+          retention-days: 14
+
+  sanitizer-lite:
+    name: Sanitizer lite (60s soak)
+    runs-on: ubuntu-22.04
```
fzipi: Same, can we use 24.04 all over?

.github/workflows/ci-fast.yml
eilandert: That's possible, I noticed this repo uses 1.26/1.28 everywhere, i didn't want to break that. I'll make the change for this commit

.github/workflows/ci-deep.yml
```diff
@@ -0,0 +1,508 @@
...
+#   * memcheck AND helgrind run under tools/soak.sh, which passes
+#     --suppressions=valgrind.suppress and --error-exitcode=99 for BOTH tools.
+#   * libcoraza is a cgo Go shared lib dlopen'd at runtime; its runtime noise
+#     is suppressed wholesale in valgrind.suppress (we check OUR connector).
+
+env:
+  FORCE_JAVASCRIPT_ACTIONS_TO_NODE24: true
+  # renovate: depName=libcoraza packageName=corazawaf/libcoraza datasource=github-releases
+  LIBCORAZA_VERSION: v1.4.0
+  NGINX_VERSION: "1.28.0"
```
fzipi: Maybe using 1.30.x?

.github/workflows/ci-deep.yml
```diff
@@ -0,0 +1,508 @@
...
+        if: failure() && github.event_name == 'schedule'
+        env:
+          DISCORD_BOT_TOKEN: ${{ secrets.DISCORD_BOT_TOKEN }}
+          DISCORD_CHANNEL_BUILDS: ${{ secrets.DISCORD_CHANNEL_BUILDS }}
+        run: |
+          [ -n "$DISCORD_BOT_TOKEN" ] || exit 0
+          curl -fsS -X POST -H "Authorization: Bot $DISCORD_BOT_TOKEN" \
+            -H "Content-Type: application/json" \
+            "https://discord.com/api/v10/channels/$DISCORD_CHANNEL_BUILDS/messages" \
+            -d "{\"content\":\"🔴 ${{ github.repository }} — CI Deep / fuzz FAILED. ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}\"}" || true
```
fzipi: Hmmmm..... what is this about? Which discord is this one? I don't think this is official.

.github/workflows/ci-deep.yml
```diff
@@ -0,0 +1,508 @@
...
+        if: failure() && github.event_name == 'schedule'
+        env:
+          DISCORD_BOT_TOKEN: ${{ secrets.DISCORD_BOT_TOKEN }}
+          DISCORD_CHANNEL_BUILDS: ${{ secrets.DISCORD_CHANNEL_BUILDS }}
+        run: |
+          [ -n "$DISCORD_BOT_TOKEN" ] || exit 0
+          curl -fsS -X POST -H "Authorization: Bot $DISCORD_BOT_TOKEN" \
+            -H "Content-Type: application/json" \
+            "https://discord.com/api/v10/channels/$DISCORD_CHANNEL_BUILDS/messages" \
+            -d "{\"content\":\"🔴 ${{ github.repository }} — CI Deep / memcheck FAILED. ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}\"}" || true
```
fzipi: Same. Discord?

.github/workflows/ci-deep.yml
```diff
@@ -0,0 +1,508 @@
...
+        if: failure() && github.event_name == 'schedule'
+        env:
+          DISCORD_BOT_TOKEN: ${{ secrets.DISCORD_BOT_TOKEN }}
+          DISCORD_CHANNEL_BUILDS: ${{ secrets.DISCORD_CHANNEL_BUILDS }}
+        run: |
+          [ -n "$DISCORD_BOT_TOKEN" ] || exit 0
+          curl -fsS -X POST -H "Authorization: Bot $DISCORD_BOT_TOKEN" \
+            -H "Content-Type: application/json" \
+            "https://discord.com/api/v10/channels/$DISCORD_CHANNEL_BUILDS/messages" \
+            -d "{\"content\":\"🔴 ${{ github.repository }} — CI Deep / helgrind FAILED. ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}\"}" || true
```
fzipi: Now that I'm seeing this one again, why not creating a new local action that we call, and feels like more DRY. Also: Discord?

.github/workflows/package.yml
```diff
@@ -17,6 +17,8 @@ jobs:
     env:
       # renovate: depName=libcoraza packageName=corazawaf/libcoraza datasource=github-releases
       LIBCORAZA_VERSION: v1.4.0
+      CCACHE_DIR: ${{ github.workspace }}/.ccache
+      CCACHE_MAXSIZE: 500M
     strategy:
       matrix:
         nginx_version: ['1.24.0', '1.26.0']
```
fzipi: Why are we using these old nginx versions? 🤔

fzipi (changes requested):
Can we also update all actions to latest version sha?

.github/workflows/ci-deep.yml
eilandert: oh crap thats my internal discord, should not be in here.  thanks for catching

.github/workflows/package.yml
eilandert: want me to bump them as well?

.github/workflows/ci-deep.yml
eilandert: forget the discord :)  ci-deep comes from our internal repo's which gets fuzzed every week by and shout on our discord when it fails.  I removed it.

ppomes:
@fzipi the pinning is done — all actions across the 6 workflows are now pinned to release SHAs (checkout, cache, setup-go, upload/download-artifact, plus the third-party cosign-installer, action-gh-release, release-please and stale). Also moved runners to ubuntu-24.04 and pinned nginx 1.30.3 with sha256 per your feedback. Could you re-review when you get a chance?

fzipi (approved):
LGTM

PR #62, opened 2026-07-06
fix: require libcoraza 1.4 ABI

eilandert:

**Severity:** Low

## What
Pin the runtime dependency to `libcoraza1 (>= 1.4.0)` in `debian/control`.

## Why
The connector requires the `coraza_is_response_body_processable` symbol, which only exists in libcoraza 1.4+. With a bare `libcoraza1` dependency the package installs against 1.3 and the module fails at dlopen time with an unresolved symbol instead of a clean dependency error.

## Testing
Adds a static contract test that keeps `debian/control` and the dlopen boundary in `ngx_http_coraza_dl.c` in sync.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the module’s dependency to require `libcoraza1` version **>= 1.4.0**.
  * Clarified that response-body processing support is **required** for supported versions.

* **Tests**
  * Added a contract test to ensure package metadata and loader expectations remain consistent with the supported runtime version.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -11,7 +11,7 @@ Rules-Requires-Root: no
 
 Package: libnginx-mod-http-coraza
 Architecture: any
-Depends: ${misc:Depends}, ${shlibs:Depends}, libcoraza1
+Depends: ${misc:Depends}, ${shlibs:Depends}, libcoraza1 (>= 1.4.0)
 Description: Coraza WAF connector module for nginx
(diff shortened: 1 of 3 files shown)


ppomes:
@coderabbitai review

ppomes:
@coderabbitai review

ppomes:
@coderabbitai review

ppomes (approved):

PR #63, opened 2026-07-06
fix: drop unused Coraza dl symbols

eilandert:
Part of an 18-PR hardening series (numbered 01–18); each PR is independent against main, recommended review order is numeric.

**Severity:** Low

## What
Remove the `coraza_rules_merge` and `coraza_add_get_args` wrappers and their dlopen plumbing.

## Why
Neither symbol is called anywhere in the connector — they are dead weight in the dlopen table, and every extra required symbol is one more way for module load to fail against a differently-built libcoraza.

## Testing
Static contract test asserts the dlopen table only contains symbols the connector actually calls.

## Summary by CodeRabbit

* **Bug Fixes**
  * Removed support for two outdated Coraza integration symbols, simplifying dynamic loading and avoiding reliance on unavailable APIs.
* **Tests**
  * Added a contract test to verify the dynamic-loading boundary no longer references the removed symbols and related wrappers.

PR #64, opened 2026-07-06
fix: keep libcoraza loaded for worker lifetime

eilandert:

**Severity:** Medium

## What
Stop calling `dynlib_close()` on libcoraza; leave the library loaded for the process lifetime.

## Why
libcoraza embeds a Go runtime. A Go runtime cannot be safely unloaded via `dlclose()` — its signal handlers, GC threads and TLS blocks stay registered, so unloading the .so leaves dangling pointers and crashes on worker shutdown/reload paths. Keeping the handle for process lifetime is the documented-safe pattern for cgo libraries.

## Testing
Static contract test asserts the close path does not call `dynlib_close`.

### DIFF
--- a/src/ngx_http_coraza_dl.c
+++ b/src/ngx_http_coraza_dl.c
@@ -166,18 +166,16 @@ ngx_http_coraza_dl_open(ngx_log_t *log)
 }
 
 /* ------------------------------------------------------------------ */
-/* Public: unload libcoraza.so                                         */
+/* Public: keep libcoraza.so loaded until the process exits            */
 /* ------------------------------------------------------------------ */
 
 void
 ngx_http_coraza_dl_close(ngx_log_t *log)
 {
     if (dl_handle != NULL) {
-        dynlib_close(dl_handle);
-        dl_handle = NULL;
-        ngx_log_error(NGX_LOG_NOTICE, log, 0,
-                      "coraza: %s unloaded",
-                      CORAZA_DYNLIB_BASENAME DYNLIB_EXT);
+        ngx_log_debug1(NGX_LOG_DEBUG_CORE, log, 0,
+                       "coraza: %s left loaded for process lifetime",
+                       CORAZA_DYNLIB_BASENAME DYNLIB_EXT);
     }
 }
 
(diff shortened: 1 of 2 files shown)


ppomes (dismissed):

ppomes:
@coderabbitai review

ppomes:
Correction to my earlier note — I did **not** push to your branch, decided to leave that to you.

The conflict here is trivial though, in case it saves you time: `dl.c` auto-merges cleanly against main. The **only** conflict is an add/add on `t/coraza-dynlib-contract.t`, because #62 (now merged) also added a file with that name. After rebasing, just combine the two assertion sets into one file — single `$dl` slurp, keep #62's ABI checks plus the close-path checks here. This resolves it (all 7 pass locally):

```perl
my $control = slurp("$root/debian/control");
my $dl = slurp("$root/src/ngx_http_coraza_dl.c");

like($control, qr/\blibcoraza1\s+\(>=\s*1\.4\.0\)/,
    'Debian package pins the libcoraza runtime ABI');
unlike($dl, qr/Optional.*coraza_is_response_body_processable/s,
    'response-body helper is not documented as optional');
unlike($dl, qr/libcoraza\s*<\s*1\.4/,
    'loader comments do not advertise pre-1.4 compatibility');
like($dl, qr/DL_SYM\(dl_is_response_body_processable,\s*coraza_is_response_body_processable\)/s,
    'response-body helper is resolved as a required symbol');

my ($close_body) = $dl =~
    qr/ngx_http_coraza_dl_close\s*\([^)]*\)\s*\{(.*?)\n\}/s;
ok(defined $close_body, 'found ngx_http_coraza_dl_close body');
unlike($close_body, qr/\bdynlib_close\b/,
    'worker exit does not unload the Go-backed libcoraza handle');
unlike($close_body, qr/\bdl_handle\s*=\s*NULL\b/,
    'worker exit keeps the loaded libcoraza handle intact');
```

Once that's in it's mergeable (and already approved). Same add/add on the test file applies to #63.

ppomes:
Merged current main in to clear the conflict — same `coraza-dynlib-contract.t` add/add as #63; combined the ABI checks (#62) with the close-path checks here, all 7 pass. Non-force update, your commits untouched.

fzipi (approved):

PR #65, opened 2026-07-06
fix: remove stale sanity check build traps

eilandert:

**Severity:** Low

## What
Remove the `CORAZA_SANITY_CHECKS` blocks (inherited from ModSecurity-nginx); keep the macro as a documented no-op for old debug builds.

## Why
The stored-header "sanity check" duplicates every response header into a per-request list that is never read by anything in the Coraza connector — it was only consumed by ModSecurity's debug tooling that never got ported. Compiling with the flag set silently costs memory per request and misleads readers into thinking the data is used.

## Testing
Module builds cleanly; full `t/` suite passes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Adjusted response-header handling to remove unnecessary debug-only side effects from standard builds.
  * Updated configuration merging so the debug-related sanity-check flag is no longer carried over to child configurations.
* **Tests**
  * Added a regression test that verifies the debug macro remains accepted and the module sources still compile successfully during builds.

### DIFF
--- a/src/ddebug.h
+++ b/src/ddebug.h
@@ -25,15 +25,7 @@
  * #define CORAZA_DDEBUG 1
  */
 
-/*
- * Setting CORAZA_SANITY_CHECKS will help you in the debug process. By
- * defining CORAZA_SANITY_CHECKS a set of functions will be executed in
- * order to make sure the well behavior of CORAZA, letting you know (via
- * debug_logs) if something unexpected happens.
- *
- * If performance is not a concern, it is safe to keep it set.
- *
- */
+/* Kept as a no-op compatibility macro for old local debug builds. */
 #ifndef CORAZA_SANITY_CHECKS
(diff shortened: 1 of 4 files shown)


ppomes:
@coderabbitai review

ppomes (dismissed):

fzipi (dismissed):

ppomes:
Merged current main in to clear the `header_filter.c` conflict with the #66/#67 resolver refactor. Resolution is main's helper-based resolvers with the `CORAZA_SANITY_CHECKS` blocks stripped — exactly this PR's intent re-applied on top of #67. Net diff is unchanged (the 9 sanity blocks + ddebug.h/module.c). Non-force update, your commits untouched.

ppomes (approved):

fzipi (approved):

PR #66, opened 2026-07-06
fix: omit NUL from synthetic server header

eilandert:

**Severity:** Medium

## What
Use `sizeof(...) - 1` for the synthesized `Server` header value length.

## Why
`sizeof` on a string literal includes the trailing NUL, so the connector was handing the WAF a `Server` value with an embedded `\0`. Rules matching on `RESPONSE_HEADERS:Server` see a corrupted value, and the stored header length no longer matches what nginx actually sends on the wire.

## Testing
Adds a regression test covering the synthetic Server header value seen by the WAF.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected the reported length of the fallback `Server` header value so it no longer includes the trailing null byte.
  * This ensures the header is seen as the exact text expected by response inspection rules.

* **Tests**
  * Added coverage for response `Server` header handling, including successful matching and a blocked response when the header value matches a rule exactly.

### DIFF
--- a/src/ngx_http_coraza_header_filter.c
+++ b/src/ngx_http_coraza_header_filter.c
@@ -78,10 +78,10 @@ ngx_http_coraza_resolv_header_server(ngx_http_request_t *r, ngx_str_t name, off_
     if (r->headers_out.server == NULL) {
         if (clcf->server_tokens) {
             value.data = (u_char *)ngx_http_server_full_string;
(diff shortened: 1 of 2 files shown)


ppomes:
@coderabbitai review

ppomes:
@coderabbitai review

ppomes (approved):

fzipi (approved):

PR #67, opened 2026-07-06
fix: propagate response header insert failures

eilandert:

**Severity:** Medium

## What
Route every response-header insertion into the WAF through one checked helper, `ngx_http_coraza_add_response_header()`, and propagate failures.

## Why
`coraza_add_response_header()` can fail (cgo boundary), but the call sites ignored the return value — on failure the WAF silently evaluates phase 3/4 rules against an incomplete header set, i.e. a fail-open. Now any insert failure is logged and aborts the filter with `NGX_ERROR`.

## Testing
Adds `t/coraza-header-error-propagation.t` (static contract: all inserts confined to the checked helper, failures propagated).

## Summary by CodeRabbit

- **Bug Fixes**
  - Updated response-header insertion so failures are detected and handled immediately rather than silently ignored.
  - Standardized header resolver outcomes to consistently return success or failure, improving error reporting across common headers.

- **Tests**
  - Added a regression test that validates error propagation behavior for response-header handling.

### DIFF
--- a/src/ngx_http_coraza_header_filter.c
+++ b/src/ngx_http_coraza_header_filter.c
@@ -68,6 +68,12 @@ static ngx_int_t
 ngx_http_coraza_add_response_header(ngx_http_request_t *r,
     ngx_http_coraza_ctx_t *ctx, ngx_str_t *name, ngx_str_t *value)
 {
+    if (ctx == NULL) {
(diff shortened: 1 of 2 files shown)


ppomes:
@coderabbitai review

ppomes (approved):

fzipi (approved):

PR #69, opened 2026-07-06
fix: propagate delayed header filter status

eilandert:

**Severity:** Medium

## What
Propagate the real return code from the request-header forwarding loop instead of collapsing everything except `NGX_ERROR`.

## Why
The resolver can return codes other than `NGX_OK`/`NGX_ERROR` (e.g. an HTTP status from an intervention). Collapsing them lost the intervention and continued processing a request the WAF had already decided to block.

## Testing
Existing `t/` suite; blocking behaviour covered by `coraza.t` and `coraza-scoring.t`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed delayed response processing to ensure non-success return codes (including buffered/delayed flush cases) are handled correctly, while still forwarding the pending response body to avoid truncated/orphaned output.
  * Ensured the system doesn’t prematurely short-circuit when header forwarding can be deferred.

* **Tests**
  * Added regression tests covering delayed header forwarding behavior and delayed response bodies, verifying full payload delivery and matching expected body length.

### DIFF
--- a/src/ngx_http_coraza_body_filter.c
+++ b/src/ngx_http_coraza_body_filter.c
@@ -270,6 +270,20 @@ ngx_http_coraza_body_filter(ngx_http_request_t *r, ngx_chain_t *in)
                 return NGX_ERROR;
             }
 
+            /*
+             * forward_header() bottoms out in the write filter, which returns
+             * NGX_AGAIN (not just NGX_OK) whenever the headers can't be fully
+             * flushed -- e.g. a full socket buffer, or limit_rate throttling
+             * via c->write->delayed even on an empty socket.  We must NOT bail
+             * on that NGX_AGAIN: the buffered body has not entered r->out yet
+             * (the write filter only buffered the headers it was handed).  On
+             * the write retry nginx calls ngx_http_output_filter(r, NULL), so
+             * this body filter would run with in == NULL and headers_delayed
+             * already 0, fall straight through, and pending_chain would be
+             * orphaned -- headers sent, body truncated.  Hand pending_chain to
+             * the body filter unconditionally; its return value carries the
+             * NGX_AGAIN up so the retry flushes headers and body together.
+             */
             out = ctx->pending_chain;
             ctx->pending_chain = NULL;
             return ngx_http_next_body_filter(r, out);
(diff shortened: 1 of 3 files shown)


ppomes:
@coderabbitai review

ppomes:
Think this one might bite on the `NGX_AGAIN` path. `ngx_http_coraza_forward_header()` is just `return ngx_http_next_header_filter(r)`, which bottoms out in the write filter — and that returns `NGX_AGAIN` (not just `NGX_OK`/`NGX_ERROR`) whenever the headers can't be fully flushed. With `if (rc != NGX_OK) return rc;` we now bail out on that `NGX_AGAIN` *before* forwarding `ctx->pending_chain` to the body filter, and `headers_delayed` is already `0` at that point.

The problem is the buffered body never made it into `r->out` — only the headers did (write_filter only buffers what it's handed). When nginx retries the write it calls `ngx_http_output_filter(r, NULL)`, so our body filter runs with `in == NULL` and `headers_delayed == 0`, falls straight through, and `pending_chain` is never sent. Net result: headers go out, body is orphaned → truncated response / Content-Length mismatch.

The old `if (rc == NGX_ERROR)` avoided this because `NGX_AGAIN` fell through and the body still got handed to `ngx_http_next_body_filter()`, which lands it in `r->out` alongside the headers so the retry flushes both.

It's easy to write off as a rare full-socket-buffer race, but `limit_rate` triggers it deterministically — rate throttling makes the write filter return `NGX_AGAIN` via `c->write->delayed` even on an empty socket, so any inspected response on the delayed path with `limit_rate` set would truncate every time.

Would something like this keep the propagation intent without stranding the body?

```c
rc = ngx_http_coraza_forward_header(r);
if (rc == NGX_ERROR) {
    return NGX_ERROR;
}
out = ctx->pending_chain;
ctx->pending_chain = NULL;
return ngx_http_next_body_filter(r, out);
```

(the body filter's return value already carries the `NGX_AGAIN` up correctly). Might be worth a regression test with `limit_rate` + `SecResponseBodyAccess On` to lock it in.

ppomes (approved):

PR #70, opened 2026-07-06
fix: remove linker path from include flags

eilandert:

**Severity:** Low

## What
Drop the hardcoded `-L/usr/local/lib` from `config`.

## Why
The connector dlopens libcoraza at runtime — nothing links against it at build time, so the flag is dead. Worse, it silently steers the linker of anything else in the build toward `/usr/local/lib` ahead of the system paths, which breaks reproducible/pbuilder builds.

## Testing
Module configures and builds without the flag; full `t/` suite passes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated build configuration so module include/link settings no longer inject linker-style `-L` search paths.
  * Kept existing dynamic linking and module build ordering unchanged (including `-ldl`).

* **Tests**
  * Added a regression test that statically checks the top-level `config` file to confirm `ngx_module_incs` excludes `-L` entries and `ngx_module_libs` includes `-ldl`.

### DIFF
--- a/config
+++ b/config
@@ -36,7 +36,6 @@ if test -n "$ngx_module_link"; then
 
 	ngx_module_deps="$ngx_addon_dir/src/ddebug.h";
     ngx_module_libs="-ldl"
-    ngx_module_incs="-L/usr/local/lib"
 
     ngx_module_order="ngx_http_chunked_filter_module \
                       ngx_http_v2_filter_module \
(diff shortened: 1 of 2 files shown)


ppomes:
@coderabbitai review

ppomes:
@coderabbitai review

fzipi (approved):

fzipi:
@eilandert Can you gpg sign this commit?

PR #71, opened 2026-07-06
fix: cap delayed response-body buffering to bound worker memory

eilandert:

**Severity:** High

## What
Track the byte count buffered for the delayed-headers feature and, once it exceeds `NGX_HTTP_CORAZA_MAX_DELAYED_BODY` (1 MiB default, build-time overridable), stop delaying: flush the delayed headers plus everything accumulated so far and let the remainder stream through.

Additionally, handle a late (post-flush) disruptive body intervention: `ngx_http_coraza_process_intervention()` returns `NGX_ERROR` once `r->header_sent` is set, and the body filter previously fell through on that error and kept streaming the denied response. The negative return is now handled explicitly — streaming stops (`NGX_ERROR`) so the intervention still terminates the transfer promptly.

## Why
While the header filter delays response headers for phase-4 inspection, the body filter accumulates every response buffer into the request pool (`pending_chain`). For a large download, or an open-ended streaming response (SSE / long-poll) that never emits `last_buf`, that buffering grows without limit and can OOM/DoS the worker. `SecResponseBodyLimit` does not help — the connector buffers for the delayed-headers feature independently of Coraza's own inspection.

Trade-off: past the cap a phase-4 rule can no longer return a clean error page for that response (an intervention degrades to a stopped transfer) — an acceptable price to avoid unbounded memory growth on oversized/streaming responses. The post-flush intervention handling above keeps that degradation prompt and fail-closed instead of streaming the rest of a denied response.

## Testing
`t/coraza-response-body-cap.t` drives a 4 MiB delayed response and asserts it completes with 200 + intact body and that the cap path logged the early flush. `t/coraza-response-body-cap-intervention.t` adds a late (post-flush) `SecResponseBodyLimit Reject` intervention and asserts the response starts (200, headers flushed early) but stops near the Coraza limit instead of streaming the full body.

## Summary by CodeRabbit

* **New Features**
  * Added a configurable limit on how much response body can be buffered while delaying headers, with a default cap of 1 MiB.

* **Bug Fixes**
  * Improved handling of intervention errors during ongoing response processing so failures surface immediately and consistently.
  * Enhanced delayed-response buffering accounting and enforce the delayed-body cap by flushing delayed headers early and stopping further buffering once exceeded.

* **Tests**
  * Added integration tests covering delayed-body cap behavior and ensuring late post-body-cap interventions are handled correctly.

### DIFF
--- a/src/ngx_http_coraza_body_filter.c
+++ b/src/ngx_http_coraza_body_filter.c
@@ -142,6 +142,13 @@ ngx_http_coraza_body_filter(ngx_http_request_t *r, ngx_chain_t *in)
                 }
                 return ngx_http_filter_finalize_request(r,
                     &ngx_http_coraza_module, ret);
+            } else if (ret < 0) {
+                ctx->intervention_triggered = 1;
+                if (ctx->headers_delayed) {
+                    ctx->headers_delayed = 0;
+                    return NGX_HTTP_INTERNAL_SERVER_ERROR;
+                }
+                return NGX_ERROR;
             }
         }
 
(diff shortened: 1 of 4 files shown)


ppomes:
@coderabbitai review

ppomes (approved):

PR #72, opened 2026-07-06
fix: don't delay HEAD response headers

eilandert:

**Severity:** Medium

## What
Never delay response headers for HEAD requests: check `r->method` explicitly in addition to `r->header_only`.

## Why
The delayed-header logic holds headers until response-body inspection finishes. A HEAD response has no body, so nothing ever arrives to release the delay — the old check relied on `r->header_only`, which nginx's final header filter sets, but module filters can run before that. Checking the method directly guarantees a delayed HEAD can never stall.

## Testing
Adds a regression test covering HEAD against a body-inspected location.

> Note: supersedes the corresponding part of #49 (closed) with a narrower, standalone fix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of `HEAD` responses so headers are forwarded without getting delayed, preventing requests from stalling in this case.
  * Ensured `HEAD` requests return the expected status while avoiding unintended response body content.

* **Tests**
  * Added coverage for `HEAD` request behavior to verify the response is successful and no body is included.

### DIFF
--- a/src/ngx_http_coraza_header_filter.c
+++ b/src/ngx_http_coraza_header_filter.c
@@ -506,8 +506,10 @@ ngx_http_coraza_header_filter(ngx_http_request_t *r)
      * page because the original 200 headers have not yet been sent to the
      * client.
      *
-     * We skip the delay for HEAD requests (no body to inspect), error pages
-     * (already an error response), and subrequests (handled independently).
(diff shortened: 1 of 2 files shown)


ppomes:
@coderabbitai review

ppomes:
@coderabbitai review

fzipi (approved):
LGTM

PR #73, opened 2026-07-06
perf: make response body header delay configurable

eilandert:

**Severity:** Low

## What
New directive `coraza_response_body_delay on|off` (default `on`, preserving current behaviour).

## Why
To give phase-4 rules a chance to replace a denied response with a clean error page, the connector delays response headers until body inspection finishes — costing TTFB on every inspected response. Operators whose loaded ruleset has no phase-4 response rules pay that latency for nothing. libcoraza exposes no phase-rule introspection, so an explicit operator toggle is the only safe way to skip the delay. With `off`, phase-3/4 interventions still stop the stream mid-flight; only the clean-error-page guarantee is forfeited.

## Testing
Adds a regression test covering both directive states.

> Note: supersedes the corresponding part of #49 (closed) with a narrower, standalone fix.

## Summary by CodeRabbit

* **New Features**
  * Added the `coraza_delay_response_headers` directive to control whether response headers are forwarded immediately or delayed until phase-4 response-body inspection.
  * Delayed header forwarding remains enabled by default and supports configuration inheritance across nested locations.
* **Bug Fixes**
  * Maintains correct phase-4 blocking behavior, including clean handling for interventions that do not produce a response body.
* **Tests**
  * Added `coraza-delay-response-headers` coverage for off/default/on modes and verified phase-4 non-body behavior remains consistent.

### DIFF
--- a/src/ngx_http_coraza_common.h
+++ b/src/ngx_http_coraza_common.h
@@ -106,6 +106,7 @@ typedef struct {
 
     ngx_flag_t                 enable;
     ngx_flag_t                 has_rules;
+    ngx_flag_t                 delay_response_headers;
 
     ngx_http_complex_value_t  *transaction_id;
 } ngx_http_coraza_conf_t;
--- a/src/ngx_http_coraza_header_filter.c
+++ b/src/ngx_http_coraza_header_filter.c
@@ -335,6 +335,7 @@ ngx_http_coraza_header_filter(ngx_http_request_t *r)
     int ret = 0;
     ngx_uint_t status;
     char *http_response_ver;
+    ngx_http_coraza_conf_t *mcf;
 
 
     /* 304 Not Modified responses are still processed for header inspection
@@ -347,6 +348,8 @@ ngx_http_coraza_header_filter(ngx_http_request_t *r)
     {
         return ngx_http_next_header_filter(r);
     }
+
+    mcf = ngx_http_get_module_loc_conf(r, ngx_http_coraza_module);
 
     if (ctx->intervention_triggered) {
         return ngx_http_next_header_filter(r);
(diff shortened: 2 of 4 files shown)


ppomes:
@coderabbitai review

ppomes:
@coderabbitai review

eilandert:
Addressed the timing-flakiness nitpick in \`timed_get\`. Replaced the fixed 0.5s early-read window + 1.5s daemon sleep with a marker handshake so there is no wall-clock racing:

- The daemon writes the \`headers-<key>\` marker after sending response headers, then **blocks on a \`release-<key>\` marker** before sending the body (instead of sleeping 1.5s).
- The client drains the socket until it is idle for a full poll interval **while the body is provably still withheld**, then writes \`release-<key>\` to let the body flow. A slow header-forward under CI load still lands in `$early`, and the body can never race into the early read.

ppomes:
@coderabbitai review

ppomes:
@eilandert the directive is handy — being able to opt out when you've got no phase-4 response rules is a real win.

One naming thought before this lands, since a public directive name is hard to change once it's in a release: `coraza_response_body_delay` actually delays the **response headers** (holding them until phase-4 finishes), not the body. Reading `coraza_response_body_delay off` I'd half-expect it to disable response *body* handling/inspection rather than "stream headers immediately." Something like `coraza_delay_headers` / `coraza_hold_response_headers` might read closer to what it does — though I get that the delay is conceptually tied to response-body processing, so it's a judgement call. No strong opinion on the exact name, just flagging it while it's still cheap to rename.

Everything else looks solid — the marker-based test sync is really clean.

eilandert:
Maybe something like this?:
When keeping default on → coraza_delay_response_headers.
When setting default to off → coraza_early_response_headers.

ppomes:
I'd go with `coraza_delay_response_headers`, default on. The feature is the delay, so a directive named after what it does and on by default is the most intuitive — `off` to opt out reads cleanly. `coraza_early_response_headers` defaulting off works too, but "early headers off = delay" is a bit of a double negative to reason about. And keeping default on means no behavior change for anyone already relying on the clean phase-4 error page.

eilandert:
Done — renamed the directive to `coraza_delay_response_headers` (default `on`) as suggested. Named after what it does, defaults on, so current behaviour and backward-compat are preserved with no double-negative. Struct field and the test file were renamed to match; module builds clean and the delay test passes (10/10). Force-pushed to this branch.

ppomes:
@fzipi @eilandert — merged current main in to clear the `header_filter.c` conflict, so this is ready for review.

The conflict was in the delay condition: main had gained `r->method != NGX_HTTP_HEAD` (from #72, merged since) and this PR adds the `mcf->delay_response_headers` gate. Combined both:

```c
if (mcf->delay_response_headers
    && r->method != NGX_HTTP_HEAD && !r->header_only && !r->error_page
    && r == r->main
    && r->headers_out.status != NGX_HTTP_SWITCHING_PROTOCOLS)
```

(and merged the two comment halves). The rename to `coraza_delay_response_headers` is intact; net change vs the pre-conflict PR is unchanged. Non-force update.

Since I authored the merge resolution, it needs an approving review from someone other than me — not a self-approve.

eilandert:
can't help you there :-)

fzipi (approved):

PR #74, opened 2026-07-06
perf: avoid copying uninspected file response buffers

eilandert:

**Severity:** Low

## What
When response-body inspection is off for a transaction (`SecResponseBodyAccess Off` / non-matching MIME type), pass non-temp file buffers through as file-range clones instead of reading them into a pool copy.

## Why
The filter memcpy'd every response buffer into `r->pool` even when the WAF never looks at the body. For static-file responses this doubles memory traffic and defeats nginx's zero-copy file path. In-memory buffers still get a pool copy (they may point into the reusable upstream buffer), and temp-file buffers keep the old path.

## Testing
Adds a regression test; existing response-body tests cover the inspected path.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of delayed, file-backed response buffers to avoid unnecessary body copying when response inspection is disabled.
  * Preserved correct delivery behavior for large file responses, including proper blocking without unintended content exposure.
* **Tests**
  * Added a new nginx-based test covering delayed file-buffer scenarios to verify both pass and deny outcomes and that delayed bodies are handled correctly.

### DIFF
--- a/src/ngx_http_coraza_body_filter.c
+++ b/src/ngx_http_coraza_body_filter.c
@@ -205,36 +205,58 @@ ngx_http_coraza_body_filter(ngx_http_request_t *r, ngx_chain_t *in)
                 u_char    *data;
                 size_t     len;
 
-                if (ngx_http_coraza_read_body_data(r, chain->buf, &data, &len)
-                    != NGX_OK)
+                if (!ctx->response_body_processable
+                    && !ngx_buf_in_memory(chain->buf)
+                    && chain->buf->in_file
+                    && !chain->buf->temp_file)
                 {
-                    return NGX_ERROR;
-                }
-
-                b = ngx_calloc_buf(r->pool);
-                if (b == NULL) {
-                    return NGX_ERROR;
-                }
-
-                if (len > 0) {
                     /*
-                     * Always make a pool copy: in-memory buffers may point
-                     * into nginx's single reusable upstream buffer (u->buffer)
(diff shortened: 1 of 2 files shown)


ppomes:
@coderabbitai review

fzipi (dismissed):

fzipi:
@copilot Verify each finding against current code. Fix only still-valid issues, skip the
rest with a brief reason, keep changes minimal, and validate.

In `@src/ngx_http_coraza_body_filter.c` around lines 208 - 211, The fast-path
check in ngx_http_coraza_body_filter.c is now broader than the body reader’s
file-buffer validity rule. Update the condition in the response-body replay path
to require the same file-backed invariants used by
ngx_http_coraza_read_body_data(), specifically that the buffer is in_file and
has a valid file pointer, before treating it as replayable. If nginx guarantees
this state elsewhere, make that guarantee explicit in the check or surrounding
logic using chain->buf and the existing body-processing path.

ppomes:
Merged current main into the branch to clear the `body_filter.c` conflict with #71 (the delayed-buffer cap). The interleave: #71's `ctx->pending_bytes += len` goes in the **copy** branch only — the clone fast-path here references the on-disk file range instead of copying into `r->pool`, so it correctly doesn't count toward the memory the cap bounds. Left a short comment in the code noting that. Non-force update (merge commit), your commits are untouched — CI should run clean now.

fzipi (approved):

PR #75, opened 2026-07-06
fix: clear entity headers on synthesized redirects

eilandert:

**Severity:** Low

## What
When a phase-3 rule turns an already-populated 200 response into a redirect (`redirect:` action), clear the full entity/representation header set of the discarded response, not just `Content-Length`.

## Why
The header filter synthesizes the `Location` and status but left `Content-Type`, `Content-Encoding`, `Last-Modified`, `ETag` and `Accept-Ranges` describing a representation the redirect no longer sends — protocol-inconsistent per RFC 9110 §15.4 / §8.3–8.8.

## Testing
Adds `t/coraza-redirect-entity-headers.t`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved redirect responses so they no longer carry over body-related headers from the original response.
  * Ensured redirects keep only the appropriate `Location` header and omit stale metadata like content type, length, last modified, ETag, and range information.

* **Tests**
  * Added coverage for redirect responses to verify that representation headers are properly cleared.

### DIFF
--- a/src/ngx_http_coraza_header_filter.c
+++ b/src/ngx_http_coraza_header_filter.c
@@ -465,13 +465,35 @@ ngx_http_coraza_header_filter(ngx_http_request_t *r)
             r->err_status = 0;
             r->header_only = 1;
 
-            /* Clear entity headers from the original response to avoid
-             * protocol-inconsistent redirects (e.g. 3xx with Content-Length
-             * but no body). */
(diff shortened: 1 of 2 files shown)


ppomes:
@coderabbitai review

ppomes (approved):

PR #76, opened 2026-07-06
fix: do not synthesize Connection/Keep-Alive for the WAF on HTTP/2

eilandert:

**Severity:** Low

## What
Skip synthesizing `Connection` / `Keep-Alive` headers for WAF inspection when `r->stream` is set (HTTP/2).

## Why
The response-header resolver synthesizes a `Connection` header (and `Keep-Alive` when `keepalive_timeout` carries a header timeout) and feeds it to Coraza. HTTP/2 forbids connection-specific header fields (RFC 9113 §8.2.2) and nginx never emits them on an h2 stream, so the WAF inspects a header the client never receives — a rule on `RESPONSE_HEADERS:Connection` would false-positive on every h2 response. `Transfer-Encoding` is already gated on `r->chunked`, which nginx never sets on an h2 stream.

## Testing
Adds `t/coraza-h2-forbidden-headers.t`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved HTTP/2+ response handling to avoid synthesizing or exposing connection-related headers such as `Connection`/`Keep-Alive` on stream-based responses.
  * This reduces cases where the WAF could observe these hop-by-hop headers inappropriately.

* **Tests**
  * Added a new HTTP/2-specific test covering forbidden response header hygiene.
  * Verifies `RESPONSE_HEADERS:Connection` triggers under HTTP/1.1 while HTTP/2 correctly keeps the synthetic `Connection` header from being visible to policy evaluation.

### DIFF
--- a/src/ngx_http_coraza_header_filter.c
+++ b/src/ngx_http_coraza_header_filter.c
@@ -222,6 +222,20 @@ ngx_http_coraza_resolv_header_connection(ngx_http_request_t *r, ngx_str_t name, 
     clcf = ngx_http_get_module_loc_conf(r, ngx_http_core_module);
     ctx = ngx_http_get_module_ctx(r, ngx_http_coraza_module);
 
+#if (NGX_HTTP_V2)
+    if (r->stream) {
+        /*
+         * HTTP/2 carries no Connection or Keep-Alive header: RFC 9113 §8.2.2
+         * forbids these connection-specific header fields and nginx never
+         * emits them on an h2 stream.  Synthesizing a phantom
+         * Connection/Keep-Alive here would make the WAF inspect a header the
+         * client never receives -- e.g. a rule on RESPONSE_HEADERS:Connection
+         * would false-positive on every HTTP/2 response.
+         */
+        return NGX_OK;
+    }
+#endif
+
     if (r->headers_out.status == NGX_HTTP_SWITCHING_PROTOCOLS) {
         connection = "upgrade";
     } else if (r->keepalive) {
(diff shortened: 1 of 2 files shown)


ppomes:
@coderabbitai review

ppomes:
@coderabbitai review

ppomes:
@eilandert nice catch on the h2 phantom-Connection — the false-positive-on-every-response reasoning is spot on.

One extension worth considering: HTTP/3 forbids the same connection-specific headers (RFC 9114 §4.2) and nginx doesn't emit them on h3 either, but `r->stream` is h2-only (it's NULL for h3 — those go through `v3_parse`), so this guard won't fire for h3 and the phantom `Connection` would still false-positive on every HTTP/3 response.

Could sidestep both with a version check instead of `r->stream`:

```c
if (r->http_version >= NGX_HTTP_VERSION_20) {   /* h2 = 2000, h3 = 3000 */
    return NGX_OK;
}
```

covers h2 + h3 in one go, and `NGX_HTTP_VERSION_20` is always defined so you don't even need the `#if (NGX_HTTP_V2)` guard. The h1/h2 test is great — an h3 case would be the natural addition if the test harness supports it.

eilandert:
Good call — pushed cc38a6b. Swapped `r->stream` for `r->http_version >= NGX_HTTP_VERSION_20`, which covers h2 (2000) and h3 (3000) in one check and drops the `#if (NGX_HTTP_V2)` guard since `NGX_HTTP_VERSION_20` is always defined. The h1-vs-h2 test still exercises the version-guard path; I left h3 without its own case since the harness has no QUIC/TLS scaffolding, but noted in the test header that h3 shares the identical path.

ppomes (approved):

PR #77, opened 2026-07-06
harden: sanitize control bytes in synthesized Location header

eilandert:

**Severity:** Low

## What
Truncate the synthesized `Location` header at the first C0 control character or DEL before copying `intervention->data` into the response; log a warning when the guard fires.

## Why
`ngx_http_coraza_process_intervention` copies `intervention->data` verbatim into `Location` on a redirect. nginx does not sanitize outgoing header values, so a CR/LF in that data would be emitted raw — HTTP response splitting / header injection. A legitimate URL carries no unencoded control bytes.

**Severity note:** defense-in-depth. libcoraza 1.4 does NOT macro-expand the `redirect:` target (verified: a `%{ARGS.x}` target reaches `Location` literally), so `intervention->data` is the static rule-author string and client data cannot reach it through a normal rule today. The guard future-proofs against a libcoraza that gains redirect macro expansion (upstream Coraza supports it) or any other dynamic source, at negligible cost.

## Testing
Adds `t/coraza-location-sanitize.t` (a clean redirect target passes through byte-for-byte, confirming no over-truncation).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved redirect handling by sanitizing outgoing `Location` header values to prevent control bytes from being included in redirect targets.
  * When unsafe bytes are detected, the redirect target is truncated and a warning is logged to keep headers safe.

* **Tests**
  * Added a regression test to verify 302 redirect responses return the expected `Location` header without unintended truncation/sanitization effects.

### DIFF
--- a/src/ngx_http_coraza_module.c
+++ b/src/ngx_http_coraza_module.c
@@ -91,6 +91,30 @@ ngx_http_coraza_process_intervention(coraza_transaction_t transaction, ngx_http_
 			if (h != NULL)
 			{
 				size_t len = ngx_strlen(intervention->data);
+				/*
+				 * Defend against response splitting / header injection.
+				 * If a rule builds the redirect target from
+				 * client-controlled data (macro expansion of a request
+				 * variable), intervention->data may contain CR/LF or other
+				 * control bytes.  nginx does not sanitize outgoing header
+				 * values, so a raw CR/LF here would let an attacker inject
+				 * extra response headers or a body.  Truncate the Location
+				 * at the first C0 control character or DEL (a legitimate
+				 * URL carries none unencoded).
+				 */
+				{
+					u_char *loc = (u_char *) intervention->data;
+					size_t safe = 0;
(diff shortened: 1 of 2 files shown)


ppomes:
@coderabbitai review

fzipi:
@copilot Verify each finding against current code. Fix only still-valid issues, skip the
rest with a brief reason, keep changes minimal, and validate.

In `@t/coraza-location-sanitize.t` around lines 11 - 16, Add a lower-level test
harness to cover the truncation branch in ngx_http_coraza_process_intervention,
since the current coraza-location-sanitize test only verifies the clean redirect
path. Create a focused test around the intervention handling path that sets
intervention->data to a string containing control bytes like CR/LF and then
asserts the emitted Location value is truncated at the first control byte. Keep
the existing rule-engine coverage intact and use
ngx_http_coraza_process_intervention as the main symbol to locate the code under
test.

ppomes (approved):

fzipi (approved):

PR #78, opened 2026-07-06
fix: guard size_t to int narrowing at the Coraza cgo boundary

eilandert:

**Severity:** Medium

## What
Add explicit fail-closed length ceilings (block / `NGX_ERROR`) at the request-side and response-body cgo call sites before narrowing `size_t` lengths to `int`.

## Why
`coraza_add_request_header`, `coraza_append_request_body` and `coraza_append_response_body` all take `int` lengths, but the connector passed `size_t` values straight through. A length above `INT_MAX` would wrap to a negative or truncated `int`, making the engine inspect the wrong span — a potential WAF bypass or over-read. nginx bounds these sizes far below `INT_MAX` in practice, so the guards are cheap and should never fire; when they do, failing closed beats silent mis-inspection.

The response-header path (`ngx_http_coraza_resolv_header_*`) is intentionally left to the header-error-propagation refactor (series PR 07), which routes those emissions through a single helper where the same guard belongs; response headers are server-generated and bounded well below `INT_MAX`.

## Testing
Adds `t/coraza-large-header-inspection.t` (a 16k request header is still inspected, confirming the guard does not clip legitimate headers).

## Summary by CodeRabbit

* **Bug Fixes**
  * Added fail-closed, overflow-safe validation for large request headers, request-body chunks, and response-body chunks before inspection/processing.
  * Oversized inputs are now rejected safely to avoid unsafe narrowing/truncation.
* **Tests**
  * Added an Nginx integration test that sends large `X-Big` headers and verifies requests containing the blocking substring are denied (403) while benign large headers are accepted (200).

### DIFF
--- a/src/ngx_http_coraza_body_filter.c
+++ b/src/ngx_http_coraza_body_filter.c
@@ -12,6 +12,26 @@
 
 static ngx_http_output_body_filter_pt ngx_http_next_body_filter;
 
+/*
+ * Length of the body data ngx_http_coraza_read_body_data would return,
+ * without performing the read/allocation -- lets callers reject an
+ * oversized chunk before paying for it.
+ */
+static size_t
+ngx_http_coraza_body_chunk_len(ngx_buf_t *buf)
+{
+    if (ngx_buf_in_memory(buf)) {
+        return buf->last - buf->pos;
+    }
+
+    if (buf->in_file && buf->file) {
+        return buf->file_last - buf->file_pos;
+    }
+
+    return 0;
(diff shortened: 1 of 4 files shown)


ppomes:
@coderabbitai review

ppomes:
@coderabbitai full review

ppomes:
@coderabbitai full review

ppomes:
@coderabbitai full review

fzipi (approved):

PR #82, opened 2026-07-23
fix: don't delay response headers for SSE streams (#81)

### LINKED ISSUE #81: Server-Sent Events never reach the client when the Coraza module is enabled (`coraza on`)
Thanks for maintaining this connector.

We use `coraza-nginx` in front of services that rely on Server-Sent Events (SSE), e.g. the ArgoCD UI (`/api/v1/stream/applications`, pod log streaming). With the module enabled (`coraza on;`), SSE responses never reach the client: no response headers, no events -> the client eventually times out. The exact same nginx config with `coraza off;` streams events immediately. This happens with the engine in `DetectionOnly` and no rules loaded beyond the engine directive, so it does not appear to be rule-related; regular request/response traffic through the same `location` works fine. Explicitly setting `SecResponseBodyAccess Off` does not change the behavior, and nginx's own response buffering is ruled out via `proxy_buffering off`.

This looks like the SSE sibling of the WebSocket issue #53.

## Environment

- `coraza-nginx` v0.11.4, built with the repo's own `Dockerfile` (nginx 1.28.3 / `nginx:stable`), libcoraza v1.6.0.
- Also tested against current `main` (`d29279f`, v0.11.5 candidate), see Observed.
- Also reproduced in production behind NGINX Ingress Controller with a custom build of this connector.

## Reproduction

`compose.yaml` (builds the connector from this repo, pinned to v0.11.4):
```yaml
services:
  sse:
    image: python:3-alpine
    command: ["python", "/sse.py"]
    volumes: ["./sse.py:/sse.py:ro"]
  waf:
    build:
      context: https://github.com/corazawaf/coraza-nginx.git#${CORAZA_REF:-v0.11.4}
      args:
        LIBCORAZA_VERSION: v1.6.0
    depends_on: [sse]
    ports: ["8080:80"]
    volumes: ["./conf.d:/etc/nginx/conf.d:ro"]
```

`sse.py` (minimal SSE backend: `/sse` emits one event per second, anything else answers a normal `200` immediately; `?pad=N` pads each event to make volume-dependent behavior observable quickly):
```python
import http.server, socketserver, time
from urllib.parse import parse_qs, urlparse

class Handler(http.server.BaseHTTPRequestHandler):
    def do_GET(self):
        if self.path.startswith("/sse"):
            pad = int(parse_qs(urlparse(self.path).query).get("pad", ["0"])[0])
            self.send_response(200)
            self.send_header("Content-Type", "text/event-stream")
            self.send_header("Cache-Control", "no-cache")
            self.end_headers()
            n = 0
            try:
                while True:
                    n += 1
                    self.wfile.write(f"data: tick {n} {'x' * pad}\n\n".encode())
                    self.wfile.flush()
                    time.sleep(1)
            except (BrokenPipeError, ConnectionResetError):
                pass
        else:
            body = b"hello\n"
            self.send_response(200)
            self.send_header("Content-Type", "text/plain")
            self.send_header("Content-Length", str(len(body)))
            self.end_headers()
            self.wfile.write(body)

    def log_message(self, *args):
        pass

socketserver.ThreadingTCPServer.allow_reuse_address = True
with socketserver.ThreadingTCPServer(("", 8000), Handler) as srv:
    srv.serve_forever()
```

`conf.d/proxy.conf`:
```nginx
coraza on;                                    # control: change to `coraza off;`
coraza_rules 'SecRuleEngine DetectionOnly';
upstream backend { server sse:8000; }
server {
  listen 80 default_server;
  location / {
    proxy_pass http://backend;
    proxy_http_version 1.1;
    proxy_buffering off;                      # rule out nginx's own response buffering
  }
}
```

Bring it up:
```bash
docker compose up -d --build
```

**Test**

```bash
# normal request: returns immediately in both modes:
curl -is --max-time 5 http://localhost:8080/hello

# SSE: expect response headers plus one "data: tick N" line per second:
curl -iN --max-time 8 http://localhost:8080/sse
```

## Observed

With `coraza-nginx` v0.11.4:

| config | `/hello` | `/sse` |
|---|---|---|
| `coraza off;` | `200`, immediate | headers + one `data: tick N` per second |
| `coraza on;`, engine `DetectionOnly` | `200`, immediate | **0 bytes received, not even response headers**; curl times out |
| `coraza on;` + `SecResponseBodyAccess Off` | `200`, immediate | unchanged, 0 bytes |

Since nothing is blocked in `DetectionOnly` (and no rules are loaded beyond the engine directive), this does not appear to be an intervention; the response simply never starts as long as the upstream keeps the stream open.

With current `main` (`d29279f`, includes the `fix: cap delayed response-body buffering to bound worker memory` change; built via `CORAZA_REF=main docker compose build waf` after making the context ref an env var):

| endpoint | result |
|---|---|
| `/sse` (~14 B/s) | still 0 bytes after 8 s, `time_starttransfer` never happens |
| `/sse?pad=131072` (~128 KiB/s) | first byte after **~7.1 s**, i.e. once ≈ 1 MiB has accumulated; streams live from then on |

So the cap bounds worker memory and eventually flushes, but a low-volume stream still stalls until 1 MiB has accumulated, for real-world SSE endpoints (the ArgoCD application-event stream emits a few KiB per minute) that is effectively still indefinite.

## Question

Would it be possible for the connector to allow streaming responses to pass through while the WAF remains enabled? For example, this could be based on the value of the SecResponseBodyAccess parameter or the Content-Type header, similar to the handling of HEAD and 101 Switching Protocols requests. We don't need the streaming bodies to be inspected; we just need the SSE endpoints to work while the WAF protects the rest of the vhost.

I'm happy to provide any further details or to test a fix against this reproduction.

eilandert:
Fixes #81.

> [!IMPORTANT]
> **Depends on #73 — please merge that first.** This branch is based on #73
> because both touch the same delay condition in `ngx_http_coraza_header_filter()`.
> Merging this one first would conflict, and the SSE term must sit alongside
> `coraza_delay_response_headers`, not replace it. Once #73 lands I'll rebase
> onto `main` if the merge isn't clean.

## The problem

The connector delays response headers until phase-4 body inspection completes at
`last_buf`. An SSE stream emits events indefinitely and never sends a `last_buf`,
so the headers are held until the delayed-body cap (1 MiB) forces a flush.

The cap bounds *memory*, not *latency*. As the reporter measured on current `main`:

| endpoint | result |
|---|---|
| `/sse` (~14 B/s) | 0 bytes after 8 s, `time_starttransfer` never happens |
| `/sse?pad=131072` (~128 KiB/s) | first byte after ~7.1 s, once ≈1 MiB accumulated |

A real SSE endpoint (ArgoCD's application-event stream emits a few KiB per minute)
reaches 1 MiB after hours, so in practice the client receives nothing.

## The fix

Skip the header delay for `Content-Type: text/event-stream`, exactly as we already
do for `101 Switching Protocols`. The media-type match is strict: only
end-of-value, or optional OWS (SP/HTAB, RFC 9110 §5.6.3) followed by `;`, qualifies.
`text/event-streamx` and `text/event-stream junk` do **not** — they keep the delay.

## Security trade-off — please read

`Content-Type` is chosen by the upstream, so **an origin emitting
`text/event-stream` opts that response out of the phase-4 header delay.**

What is *not* affected:
- Phases 1–3 are untouched.
- Phase 4 still **runs** — the body filter always calls `coraza_process_response_body()`.

What is given up: the ability to turn a phase-4 match into a *clean error page*,
because the headers are already on the wire. A late match degrades to a connection
reset instead.

This is the same trade-off `101 Switching Protocols` already accepts, and it is
inherent — streaming and full-response WAF buffering are mutually exclusive by
construction. Flagging it explicitly so the choice is deliberate rather than
discovered later.

An alternative considered and rejected: gating on `ctx->response_body_processable`
(i.e. `SecResponseBodyAccess Off`) instead of the Content-Type. That reads like the
tidier fix — the delay condition's own comment already promises it — but it is
**wrong**: phase-4 rules keyed on non-body variables (`ARGS`, `TX`) fire regardless
of body inspection, and skipping the delay for them breaks their clean error page.
Verified: it fails 5 phase-4 cases in `coraza.t` with
`[alert] header already sent while sending response to client`.

If maintainers prefer the exemption to be opt-in rather than upstream-driven, the
natural follow-up is gating it behind the #73 directive — happy to do that instead.

## Tests

`t/coraza-sse.t` — the upstream daemon never closes the connection, so a delayed
response can never be rescued by an upstream-close flush.

Controls, so this can't silently regress into a bypass:
- near-miss media types must **stay** delayed;
- a phase-1 rule must still **block** an SSE request.

Verified locally against nginx 1.28.0 (ASan/UBSan):
- **with** the fix: 12/12 pass;
- **without** it: 7 fail, reproducing #81 exactly;
- full `coraza*.t`: no regressions — `coraza.t`'s phase-4 block/redirect cases stay green.

## Summary by CodeRabbit

* **New Features**
  * Server-Sent Events responses now stream promptly without waiting for the connection to close.
  * SSE responses continue to undergo web application firewall inspection and enforcement.

* **Bug Fixes**
  * Improved media-type detection to avoid treating near-match content types as SSE responses.
  * Preserved request blocking behavior for protected SSE endpoints.

### DIFF
--- a/src/ngx_http_coraza_header_filter.c
+++ b/src/ngx_http_coraza_header_filter.c
@@ -182,6 +182,64 @@ ngx_http_coraza_resolv_header_content_type(ngx_http_request_t *r, ngx_str_t name
 }
 
 
+/*
+ * A streaming response has no meaningful end-of-body: the origin emits events
+ * indefinitely and never sends a last_buf, so the header-delay flush (which
+ * fires on last_buf) holds the response headers until the delayed-body cap
+ * (NGX_HTTP_CORAZA_MAX_DELAYED_BODY) is reached.  A real SSE endpoint emits a
+ * few bytes per event, so that cap is reached only after minutes or hours --
+ * in practice the client receives nothing (issue #81).
+ *
+ * Detect Server-Sent Events (Content-Type: text/event-stream) so the caller
+ * can skip the delay, exactly as it does for 101 Switching Protocols.
+ *
+ * SECURITY TRADE-OFF: Content-Type is chosen by the upstream, so an origin
+ * that emits text/event-stream opts this response out of the phase-4 header
+ * delay.  Phase 4 still RUNS on such a response (the body filter always calls
+ * coraza_process_response_body()), and phase 1-3 are untouched -- what is lost
+ * is only the ability to turn a phase-4 match into a clean error page, because
+ * the headers are already on the wire.  A late match degrades to a connection
+ * reset instead.  That is the same trade-off 101 Switching Protocols already
+ * accepts, and it is inherent: streaming and full-response WAF buffering are
+ * mutually exclusive by construction.  Operators who do not proxy untrusted
+ * origins and want the delay unconditionally can leave their upstreams from
+ * emitting text/event-stream, or disable streaming endpoints at the proxy.
+ */
+static ngx_int_t
+ngx_http_coraza_is_sse_response(ngx_http_request_t *r)
+{
+    ngx_str_t *ct = &r->headers_out.content_type;
+    static const u_char sse[] = "text/event-stream";
+    size_t sse_len = sizeof(sse) - 1;
+    size_t i;
+
+    if (ct->len < sse_len) {
+        return 0;
+    }
+
+    /* Match the media type; tolerate a trailing "; charset=..." etc. */
+    if (ngx_strncasecmp(ct->data, (u_char *) sse, sse_len) != 0) {
+        return 0;
+    }
+
+    /*
+     * The media type must be followed by end-of-value or a semicolon-
+     * delimited parameter list, with optional OWS (SP / HTAB, RFC 9110
+     * 5.6.3) before the semicolon.  Anything else ("text/event-streamx",
+     * "text/event-stream application/json") is NOT SSE and must keep the
+     * phase-4 header delay.
+     */
+    for (i = sse_len; i < ct->len; i++) {
+        if (ct->data[i] == ' ' || ct->data[i] == '\t') {
+            continue;
+        }
+        return ct->data[i] == ';';
+    }
+
+    return 1;
+}
+
+
 static ngx_int_t
 ngx_http_coraza_resolv_header_last_modified(ngx_http_request_t *r, ngx_str_t name, off_t offset)
 {
@@ -524,11 +582,19 @@ ngx_http_coraza_header_filter(ngx_http_request_t *r)
      * body, so the body filter never sees a last_buf to trigger the flush.
      * Delaying the 101 would hold the handshake forever and the upgrade
      * would never complete.
+     *
+     * We likewise skip Server-Sent Events (Content-Type: text/event-stream):
+     * an SSE stream emits events indefinitely and never sends a last_buf, so
+     * the headers would be held until the delayed-body cap forces a flush --
+     * minutes or hours on a real event stream, i.e. the client gets nothing
+     * (issue #81).  See ngx_http_coraza_is_sse_response() above for the
+     * security trade-off this accepts.
      */
     if (mcf->delay_response_headers
         && r->method != NGX_HTTP_HEAD && !r->header_only && !r->error_page
         && r == r->main
-        && r->headers_out.status != NGX_HTTP_SWITCHING_PROTOCOLS)
+        && r->headers_out.status != NGX_HTTP_SWITCHING_PROTOCOLS
+        && !ngx_http_coraza_is_sse_response(r))
     {
         /*
          * Delay sending headers until phase 4 completes so that
--- a/t/coraza-sse.t
+++ b/t/coraza-sse.t
@@ -1,0 +1,225 @@
+#!/usr/bin/perl
+
+# Tests for Coraza-nginx connector (Server-Sent Events, text/event-stream).
+#
+# Regression test for issue #81: the connector delays response headers until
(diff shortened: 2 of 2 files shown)


eilandert:
Pushed 250b797 addressing review findings:

- **Strict SSE media-type boundary** — after `text/event-stream`, only end-of-value or optional OWS (SP/HTAB, RFC 9110 §5.6.3) followed by `;` is accepted. Fixes both directions: legal HTAB OWS no longer falls back to the stalling delayed path, and malformed values like `text/event-stream junk` no longer get the phase-4 exemption.
- **Helper renamed** to `ngx_http_coraza_is_sse_response()` — it only detects SSE.
- **Test now actually regresses**: the upstream daemon streams heartbeats until the client disconnects and never closes on its own, so the pre-fix code can't pass via the upstream-close `last_buf` flush. RED-verified: fails 6/6 positive assertions on the parent commit, passes on this branch. Also switched to `sysread()` (`read()` blocks for the full requested length, which a trickling stream never delivers) and added variant coverage (charset param, HTAB OWS, mixed case) plus negative controls asserting near-miss content types keep the header delay.

Full local suite: 32 files / 345 tests PASS (ASan/UBSan build).

ppomes:
Hi @eilandert,

#73 is now merged, as you suggested, you can rebase the branch! 

Pierre

ppomes (approved):

PR #83, opened 2026-07-23
fix: report HTTP/3 protocol version to the WAF

eilandert:

## TL;DR
The WAF was told every HTTP/3 request and response was plain old HTTP/1.x, because the code looked at the wrong flag. Any rule that says "only for HTTP/3" (or "only HTTP/1.1") therefore misfired on QUIC traffic. Now it reads the real version, h2 and h3 included.

**Severity:** Medium

## What
Derive the protocol version reported to the WAF from `r->http_version` instead of `r->stream`, and add the missing `NGX_HTTP_VERSION_30` case to the request-side switch.

## Why
`ngx_http_coraza_header_filter.c` selected the response protocol string with:

```c
http_response_ver = "HTTP 1.1";
#if (NGX_HTTP_V2)
    if (r->stream) { http_response_ver = "HTTP 2.0"; }
#endif
```

`r->stream` is set only for HTTP/2, so **every HTTP/3 response was reported to the engine as `HTTP 1.1`**. The request side had the same gap from the other direction: its `switch (r->http_version)` has cases for 0.9/1.0/1.1/2.0 and an HTTP/3 request fell through to `default:` and was reported as `1.0`.

Any rule keyed on the request or response protocol therefore mis-fires on QUIC traffic — it either matches HTTP/3 traffic as if it were HTTP/1.x, or fails to match a rule written to target HTTP/3. Note the neighbouring `ngx_http_coraza_resolv_header_connection` already uses `r->http_version >= NGX_HTTP_VERSION_20` for exactly this reason, so the file was internally inconsistent.

Both changes are guarded on `nginx_version >= 1025000` (HTTP/3 landed in nginx 1.25.0), matching the existing `nginx_version >= 1009005` guard on the HTTP/2 case.

## Testing

Behavioural coverage needs an HTTP/3 listener, which the `t/` suite has no harness for today (it has no non-IPv4-TCP listener at all). A `REQUEST_PROTOCOL` / response-protocol assertion over h3 is queued for the test-coverage PR rather than bundled here.

## Summary by CodeRabbit

* **New Features**
  * Added HTTP/3 protocol recognition for request and response processing.
  * Coraza now reports HTTP/3 traffic using the correct protocol version.

* **Bug Fixes**
  * Improved protocol detection for HTTP/1.1 and HTTP/2 responses.

* **Tests**
  * Added coverage for HTTP/3 request and response handling.
  * Added regression checks for protocol-version mapping across supported nginx versions.
  * Expanded fuzzing coverage with additional input samples.


eilandert:
H3 can't be tested as long as #61 is not merged, the NGINX in CI is too old for H3

ppomes:
@coderabbitai full review

eilandert:
still can't test this, i suspect ubuntu 24.04 is too old and not installing a new openssl

eilandert:
Updated: the response side described in the PR body was missing from the diff — the header filter still selected the protocol string from `r->stream` (HTTP/2 only), so an HTTP/3 response was still reported as HTTP/1.1. Added the response-side fix in `ngx_http_coraza_header_filter.c` (derive from `r->http_version`, guarded on `nginx_version >= 1025000`) plus a phase-5 `RESPONSE_PROTOCOL` assertion in the h3 integration test. Also rebased on main to clear the build.yml conflict.

ppomes:
FWIW the OpenSSL version is the real gate for testing this, more than nginx. Server-side QUIC only landed in OpenSSL 3.5, so:

- Ubuntu 24.04 (current CI runner): OpenSSL 3.0.13 → nginx can't do H3 at all
- Ubuntu 26.04: OpenSSL 3.5.5 → H3 works

So testing H3 needs a runner with OpenSSL ≥ 3.5 (ubuntu-26.04 or a custom OpenSSL/quictls build), not just a newer nginx. The code itself is fine — it just can't be exercised on the 24.04 runner.

ppomes:
CI prerequisite for testing this is up in #102 — moves the runners to ubuntu-26.04 (OpenSSL 3.5, needed for nginx HTTP/3) and enables http_v3_module. Once that lands, H3 can actually be built and exercised here.

eilandert:
I added/enabled the H3 tests, ci failed, will work when #102 has been merged with main

eilandert:
sigh

| Source | Convention |
| --- | --- |
| Coraza's own middleware, `http/middleware.go:41` | `tx.ProcessURI(..., req.Proto)` — Go's `req.Proto` is `"HTTP/1.1"` |
| Coraza tests, `interceptor_test.go:630` | `ProcessURI("/", "GET", "HTTP/1.1")` |
| libcoraza tests, `tests/simple_get.c:60,66` | `"HTTP/1.1"` both request and response |
| Upstream `ProcessURI`, `transaction.go:819-825` | `requestProtocol.Set(httpVersion)` verbatim, no prefixing |
| ModSecurity-nginx `access.c:207-208` | strips `HTTP/` → bare `"1.0"` |

eilandert:
@ppomes

ppomes (approved):

PR #86, opened 2026-07-23
fix: keep transaction_id length consistent on allocation failure

eilandert:

## TL;DR
On an out-of-memory hiccup the transaction-id ended up as "empty pointer, non-zero length" — and the log line trusted the length and read the empty pointer, risking a crash. Now the length is set to match: no id present, nothing read.

**Severity:** Low

## What
Assign `ctx->transaction_id.len` from the result of `ngx_pstrdup` so the pointer and the length stay consistent when the allocation fails.

## Why
`ngx_http_coraza_create_ctx` stored the copy unchecked and set the length unconditionally:

```c
ctx->transaction_id.data = ngx_pstrdup(r->pool, &s);
ctx->transaction_id.len = s.len;
```

On allocation failure that leaves `.data == NULL` paired with a non-zero `.len`. The access-denied log line in `ngx_http_coraza_process_intervention` formats exactly that pair with `"%V"`:

```c
if (ctx->transaction_id.len > 0) {
    ngx_log_error(..., "Coraza: Access denied with code %d, unique_id \"%V\"",
        intervention->status, &ctx->transaction_id);
}
```

The guard tests `.len` only, so `%V` dereferences the NULL `.data`.

Deriving `.len` from the result treats the id as simply absent instead. Returning `NULL` here was the alternative, but the cleanup handler is not registered until further down the function, so an early return would leak the transaction that was just created — a worse trade for a path that only feeds a log line. The engine already received the id via `coraza_new_transaction_with_id()`; this copy exists solely for that message.

## Testing

The failure needs pool-allocation fault injection to reach, so there is no `t/` test; the change is a consistency fix on an error path that is unreachable under normal load.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of configured transaction IDs when memory allocation fails.
  * Prevented invalid transaction ID data from appearing in access-denied log messages.

### DIFF
--- a/src/ngx_http_coraza_module.c
+++ b/src/ngx_http_coraza_module.c
@@ -186,8 +186,17 @@ ngx_http_coraza_create_ctx(ngx_http_request_t *r)
 			return NULL;
 		}
 		ctx->coraza_transaction = coraza_new_transaction_with_id(waf, (char *)s.data);
+
+		/*
+		 * Keep transaction_id.len consistent with .data: on allocation
+		 * failure leave the id empty rather than pairing a NULL pointer
+		 * with a non-zero length.  The engine already has the id (it was
+		 * passed to coraza_new_transaction_with_id above); this copy only
+		 * feeds the "%V" in the access-denied log line, which is guarded
+		 * on .len > 0.
+		 */
 		ctx->transaction_id.data = ngx_pstrdup(r->pool, &s);
-		ctx->transaction_id.len = s.len;
+		ctx->transaction_id.len = ctx->transaction_id.data ? s.len : 0;
 	}
 	else
 	{


ppomes:
@coderabbitai full review

eilandert:
Superseded by #95 — closing as a duplicate.

Both PRs fix the same defect in `ngx_http_coraza_create_ctx()`: `ctx->transaction_id.len` was set to `s.len` unconditionally, so an `ngx_pstrdup()` failure left a `NULL` `.data` paired with a non-zero `.len`.

This PR narrows the window (`len = data ? s.len : 0`). #95 removes it entirely and fixes two adjacent problems in the same block:

- the `transaction_id` copy is moved *after* `ngx_pool_cleanup_add()`, and a failed `ngx_pstrdup()` now returns `NULL` outright — so `.data`/`.len` can never disagree, rather than agreeing on "empty";
- `coraza_new_transaction*()` returning `0` is now checked (previously a fail-open: every later `coraza_process_*`/`coraza_add_*` call would inspect nothing), and `coraza_free_transaction()` is called if the cleanup registration fails.

Nothing in this PR is missing from #95, so there is no reason to carry both through review and rebase. Continuing in #95.

PR #87, opened 2026-07-23
fix: coraza_transaction_id takes exactly one argument

eilandert:

## TL;DR
`coraza_transaction_id $a $b;` was accepted but silently used only `$a` and threw `$b` away. A config that asks for something the module can't do should be rejected up front, not quietly reinterpreted. Now it takes exactly one argument, like its two sibling directives.

**Severity:** Low

## What
Declare `coraza_transaction_id` as `NGX_CONF_TAKE1` instead of `NGX_CONF_1MORE`.

## Why
The directive accepted one-or-more arguments, but `ngx_conf_set_transaction_id` compiles only `value[1]` as its complex value:

```c
ccv.value = &value[1];
```

So `coraza_transaction_id $a $b;` passed config validation, started nginx, and silently used only `$a` — the surplus argument was accepted and discarded with no warning. A config expressing an intent the module does not implement should be rejected at parse time, not quietly reinterpreted.

`coraza_rules` and `coraza_rules_file` are both already `NGX_CONF_TAKE1` and read `value[1]` the same way, so this makes the three directives consistent.

This is a contract tightening: a config that was previously accepted now fails `nginx -t`. That config was already not doing what it said, so the failure surfaces a latent bug rather than introducing one.

## Testing
Compiles clean under `-Werror` with ASan/UBSan against nginx 1.28.0. Existing `t/coraza-transaction-id.t` passes the single-argument form and is unaffected.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the `coraza_transaction_id` configuration directive to require exactly one argument, improving validation and ensuring incorrect argument counts are rejected during configuration checks.
* **Tests**
  * Added a regression test covering valid and invalid `coraza_transaction_id` usages to confirm nginx fails fast with the expected error when the directive is misconfigured.


ppomes:
@coderabbitai full review

ppomes (approved):

PR #91, opened 2026-07-24
ci testing:  increase code coverage

eilandert:

## TL;DR
Adds the missing test cases — HTTP/0.9, many-headers, address/URI/body filtering, delayed and file-backed bodies, response-body inspection, redirects, rules-from-file — so the behaviours the other PRs fix are actually exercised in CI instead of taken on faith.

**Severity:** None (tests only — no runtime code change)

Increase testing code coverage

## Summary by CodeRabbit

- **Tests**
  - Expanded integration coverage for HTTP/0.9 handling and requests with many headers.
  - Added new checks for phase-1 filtering by client address and URI, including delayed body reads and file-backed request bodies.
  - Introduced response-body and redirect scenarios (entity header clearing, redirect status codes, error-page re-entry) plus synthesized response header behavior.
  - Added validation for external rules files, server-sent events, and safer logging behavior on malformed requests.


eilandert:
All 4 CodeRabbit findings verified against the code and addressed in the amended commit (force-pushed):

1. **redirect-clears-entity-headers.t** — added `Content-Type` and `Accept-Ranges` cleared assertions (both are cleared in `ngx_http_coraza_header_filter.c`), sent `Accept-Ranges` from the upstream fixture, and split the positive control into per-header checks so each cleanup branch is guarded. Verified by mutation: disabling either clear now fails the matching assertion.
2. **synth-response-headers.t /keepalive** — added a dedicated `/keepalive-timeout` location asserting `RESPONSE_HEADERS:Keep-Alive @streq timeout=30` (418), plus a `Connection: close` negative control. Mutation-verified: disabling the `Keep-Alive` synthesis fails test 2 while the control still passes.
3. **synth-response-headers.t SSE** — `return 200` cannot prove the delay was skipped, so added a static assertion (same source-grep idiom as the resolver branches) that the delay condition is gated on `!ngx_http_coraza_is_sse_response(r)`.
4. **synth-response-headers.t Vary/upgrade** — extended both source-grep patterns to require the subsequent `ngx_http_coraza_add_response_header` call, so they now verify WAF delivery, not just value selection.

All affected tests pass locally (nginx 1.28.0, ASan); negative controls confirmed.

eilandert:
## coverage tests + dead-code removal

Can be merged after CI changes in #90,  currently set in draft

Ran a full pass on this branch (nginx 1.31.3, `--coverage`, ASan/UBSan): **53 test files / 0 fail**, and the 13 new tests individually **65 ok / 0 not-ok, 0 sanitizer errors**.

### C change (only non-test hunk) — safe
`ngx_http_coraza_log(void*, const void*)` is removed. It was never wired: no caller and no libcoraza log-callback registration anywhere in `src/`. The real LOG-phase handler (`log_handler`) is untouched.

### Test integrity
- All 13 additions are behavioral (real requests + status/body assertions), each deny paired with a passing negative control.
- `coraza-redirect-clears-entity-headers.t` is both-sided: 6 `unlike` (headers cleared on the synthesized redirect) + 6 `like` controls (same headers kept on a clean response).
- The 4 `like($src, …)` source-greps in `coraza-synth-response-headers.t` only supplement the behavioral asserts — they pin the SSE-delay-skip / Vary / upgrade sink paths that aren't observable on the wire (the flags are set by later filters).

### Coverage
Genuine plateau — module 80.9 / rewrite 81.9 / header_filter 87.0 / body_filter 63.2 / pre_access 83.6 / log **100** / utils 87.5 / dl 95.4. Everything still uncovered needs fault injection or a quic build: alloc-fail / config-parse-fail paths, HTTP/3-only `http_version` default, and the body_filter file-read/clone branches. The file-read path was re-checked live with `sendfile off` + `directio 1` + `output_buffers 1 512` on a 200 KB static file — the buffer still arrives in-memory at the body filter (the `in_file` branch never fires), so it's defensive-but-unreachable at this filter position. Not chasing those with fake tests.

### Nits (non-blocking)
- `common.h:160` still carries the `/* ngx_http_coraza_log.c */` section comment — the file remains (holds `log_handler`), so it's just narrower now; can ride a future common.h change.

fzipi (dismissed):

fzipi:
Hey @eilandert! Can you please start gpg signing _all_ your code modifications? That will give us a solid base 😄

eilandert:
> Hey @eilandert! Can you please start gpg signing _all_ your code modifications? That will give us a solid base 😄

yes thats the idea but somehow it sometimes gets lost :(

fzipi:
> > Hey @eilandert! Can you please start gpg signing _all_ your code modifications? That will give us a solid base 😄
> 
> yes thats the idea but somehow it sometimes gets lost :(

Maybe using `git config commit.gpgsign true`?

eilandert:
> > > Hey @eilandert! Can you please start gpg signing _all_ your code modifications? That will give us a solid base 😄
> > 
> > yes thats the idea but somehow it sometimes gets lost :(
> 
> Maybe using `git config commit.gpgsign true`?

Mm that's per repo, I enabled it now. Might get lost again if I fork again 😇

ppomes (approved):

PR #94, opened 2026-07-24
fix: bound delayed-header buffering on uninspected file-backed streams

eilandert:

## TL;DR
When headers are held back for a late check, buffers pile up waiting. For file-backed streams the size counter wasn't being charged, so the safety cap that's supposed to stop unbounded memory growth never tripped. Now the file range is counted, so the cap governs it too.

**Severity:** Low–Medium (worker-memory growth / DoS on an open-ended file-backed response)

## What
When response headers are delayed for phase-4 inspection, the body filter moves each buffer into `pending_chain`. For **uninspected, non-temp, file-backed** buffers it takes a fast path (`ngx_http_coraza_body_filter.c`): clone the buffer as a file range instead of reading it into `r->pool`, and skip `pending_bytes += len` because a file range consumes no pool memory.

## Why that's a problem
`NGX_HTTP_CORAZA_MAX_DELAYED_BODY` is the **only** thing that stops the connector holding headers on an open-ended response — its comment even says it bounds "large or open-ended (streaming) responses." But on the file-range path `pending_bytes` stayed at 0, so:

- the cap at `if (ctx->pending_bytes > NGX_HTTP_CORAZA_MAX_DELAYED_BODY)` **never tripped**;
- every buffer still cost a `pending_chain` link, so a long-lived file-backed stream (repeated sendfile ranges, no `last_buf`) accumulated chain links in `r->pool` **without bound**.

Reachable because the delay is armed even when `!response_body_processable` (headers are still held; only `filter_need_in_memory` is skipped).

## Fix
Charge the logical file range (`file_last - file_pos`) to `pending_bytes` on the clone path, so the existing cap governs it too. Past the cap the headers + everything buffered so far are flushed and the remainder streams through — bounding worker memory. No cost to clean phase-4 blocking: body inspection is disabled for this transaction, so the delay was only holding headers, not enabling a body block.

## Testing
`t/coraza-delayed-file-buffer.t`, 12 assertions, all passing:

- a source-shape assertion that the clone branch charges `pending_bytes` — this is what actually pins the change;
- a runtime case (`/file-cap`, >1 MiB, headers delayed by a phase-4 rule) asserting a 200, an intact body, and the cap's early-flush warning in `error.log`.

**Coverage gap, stated plainly:** the runtime case exercises the *in-memory* path, not the file-backed clone branch this PR changes. `/file-cap` uses `sendfile off`, so nginx copies the file into memory and the cap trips via the pre-existing `pending_bytes += len` at line 320. Verified by reverting the file-range charge and re-running: the runtime assertions still pass. So they do not guard this change.

Switching to `sendfile on` does not help — `ngx_http_static_module.c:261-265` emits a static file as one buffer with `last_buf = 1`, which makes `is_request_processed` true and skips the cap check entirely (it lives in the `!is_request_processed` arm). Reaching the clone branch needs repeated non-final non-temp file buffers; the static handler cannot produce them, and upstream temp files are excluded by the `!temp_file` guard. No such producer is available in the current test build, so the branch has no runtime coverage here.

Verification performed: built nginx 1.29.0 with the module against libcoraza and ran the suite locally (12/12), plus a revert-and-rebuild negative control to establish what each assertion discriminates. Compiles clean under `-Werror`. The arithmetic reuses the existing `off_t` file range already validated against `INT_MAX` upstream in the same function.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed delayed-response buffering to correctly account for cloned, non-temporary file-backed body ranges when tracking buffered bytes.
  * Oversized delayed file-backed responses now hit the delayed-body cap more reliably and flush headers at the expected time, while still streaming the complete response body.
  * Ensured the delayed-body flush limit is consistently enforced for delayed file-stream responses as well.
* **Tests**
  * Expanded coverage with additional delayed file-buffer scenarios and stricter log/body assertions.


ppomes:
@coderabbitai full review

ppomes:
@coderabbitai full review

eilandert:
Confirmed and fixed in `18d4a2c`.

The `/file-cap` status and body-length checks did pass with the cap fix reverted, so they were guarding nothing. Added the `error.log` assertion (the warning text at `ngx_http_coraza_body_filter.c:397` matches the suggested pattern) and bumped the plan to 12.

One correction, found while checking that the assertion is a real negative control rather than assuming it: **the new assertion pins the in-memory path, not the file-backed clone branch this PR changes.**

`/file-cap` uses `sendfile off`, so nginx copies the file into memory buffers and the cap trips via the pre-existing `ctx->pending_bytes += len` path at line 320. The assertion therefore still passes with the file-range charge reverted.

Flipping the location to `sendfile on` does not fix it either — it makes the test fail outright. `ngx_http_static_module.c:261-265` emits the whole file as a single buffer with `last_buf = 1`:

```c
b->file_pos = 0;
b->file_last = of.size;
b->in_file = b->file_last ? 1 : 0;
b->last_buf = (r == r->main) ? 1 : 0;
```

With one final buffer `is_request_processed` is true, so the flow takes the end-of-body branch and the cap check — reachable only in the `!is_request_processed` arm — never runs. `output_buffers` does not apply to static files; it governs upstream reads. Exercising the clone branch needs repeated *non-final* non-temp file buffers, which the static handler cannot produce and which upstream temp files are excluded from by the `!temp_file` guard.

So the file-range charge remains pinned only by the source-shape assertion at the top of the file. That gap is recorded in a comment next to the assertion rather than left looking like runtime coverage, and I have corrected the Testing section of the PR body, which claimed the runtime case covered this path. If there is a producer of repeated non-final non-temp file buffers available in the CI build, I am happy to add real coverage for the branch.

ppomes (approved):

PR #96, opened 2026-07-24
fix: fail closed when request-body submission to Coraza fails

eilandert:

## TL;DR
nginx forwards the whole letter to the backend, but the security scanner only got to read half of it and said "couldn't finish." The old code shrugged and sent the letter anyway. An attacker could hide the nasty part in the half the scanner never saw. Now: if the scanner can't read the whole thing, we refuse the letter.

**Severity:** Medium (request-body inspection bypass — a request-smuggling-style evasion)

## What
Both request-body submission paths in the pre-access handler discarded the `int` returned by libcoraza (`ngx_http_coraza_pre_access.c`):

- `coraza_append_request_body()` — in-memory chunk path
- `coraza_request_body_from_file()` — temp-file path (previously had **no** failure handling at all)

## Why that's a problem
A failure return means the body (or a chunk) was **not** submitted to the engine, so phase-2 rules run against a **shorter body than nginx forwards upstream** — a request-smuggling-style inspection bypass. The post-append intervention poll only catches an actual interruption, not a submission failure, so the miss was silent (**fail-open**).

## Fix
Both calls now check the return and **fail closed** (500), matching how the header path already checks `coraza_add_request_header`.

> **Reviewer note (polarity):** this PR tests `< 0`. The libcoraza C export signals an append/file error with a **positive** sentinel (`return 1`), so a `< 0` test never fires — the guard should be `!= 0` to actually fail closed. See the response-side twin #100, which uses `!= 0`, and the inline comment on this PR. Recommend switching to `!= 0` before merge.

## Testing
Existing `t/coraza-request-body*.t` cover the success path. The failure path requires a libcoraza-side append error (fault injection), so there is no black-box negative control.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved request processing reliability by treating security inspection failures as errors instead of allowing requests to continue.
  - Requests that cannot be inspected correctly are now rejected with an internal server error, helping prevent unvalidated traffic from proceeding.

### DIFF
--- a/src/ngx_http_coraza_pre_access.c
+++ b/src/ngx_http_coraza_pre_access.c
@@ -137,12 +137,14 @@ ngx_http_coraza_pre_access_handler(ngx_http_request_t *r)
             dd("request body inspection: file -- %s", file_name);
 
             /*
-             * A negative return means libcoraza could not read/submit the
+             * A non-zero return means libcoraza could not read/submit the
              * body file, so the engine would evaluate phase 2 against no (or
              * a partial) body while nginx forwards the full body upstream.
              * Fail closed rather than inspect less than we forward.
+             * (libcoraza signals failure with a positive sentinel, so test
+             * != 0, not < 0.)
              */
-            if (coraza_request_body_from_file(ctx->coraza_transaction, file_name) < 0) {
+            if (coraza_request_body_from_file(ctx->coraza_transaction, file_name) != 0) {
                 ngx_log_error(NGX_LOG_ERR, r->connection->log, 0,
                     "coraza: failed to submit file-buffered request body for inspection");
                 ctx->intervention_triggered = 1;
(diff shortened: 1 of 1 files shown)


eilandert:
Heads-up while this is still open: the libcoraza C exports signal an append/submit failure with a **positive** return (`1`), not a negative one — `coraza_append_request_body` / `coraza_request_body_from_file` both `return 1` on the Go-side `WriteRequestBody(...) err != nil` / file-read error path, and `return 0` on success. So the `< 0` tests here never fire, and the fail-closed path is effectively dead.

Suggest `!= 0` instead of `< 0` (same convention used in the response-side twin, #100). Refuting welcome if the vendored libcoraza you build against differs.

eilandert:
Fixed the sentinel-polarity bug in the two fail-closed guards (commit 4408b0c): libcoraza signals `coraza_request_body_from_file` / `coraza_append_request_body` failure with a **positive** sentinel (`return 1` on error, `0` on success — verified against libcoraza 1.6.0 `coraza.go` exports), so the original `< 0` tests were dead code and the body-submission path still failed **open**. Now testing `!= 0`, so a failed submission returns 500 instead of inspecting a shorter body than nginx forwards upstream. Comments updated to state the positive-sentinel contract.

ppomes (approved):

PR #99, opened 2026-07-24
chore: code hygiene cleanup

eilandert:

## TL;DR
Wiping down the workbench: fixing a mislabelled drawer, tossing scribbled sticky-notes, removing a switch that was always on. The machine runs exactly the same after.

**Severity:** None (hygiene, no behaviour change)

## What
Non-functional tidy-ups:

- **`header_filter.c`** — `resolv_header_last_modified` used `value.len = (int)(p - buf)`, narrowing a `ptrdiff_t` through `int` into a `size_t` field; the sibling content-length resolver already uses `(size_t)`.
- **`header_filter.c`** — stale comment referencing the ModSecurity name `msc_process_response_headers()` (plus a `paramters` typo); the code calls `coraza_process_response_headers()`.
- **`header_filter.c`** — commented-out dead `content_length_n = -1` statement (the explanatory comment is kept).
- **`body_filter.c`** — an `if (!is_request_processed)` branch holding only a commented-out `dd()`.
- **`pre_access.c`** — a permanently-true `#if 1`/`#endif` wrapper left over from the ModSecurity port.

## Why that's a problem
Each is misleading rather than broken: the narrowing cast is a latent-bug smell inconsistent with its sibling, the stale comment names a function that no longer exists, and the dead `#if 1` / commented branches read as if they gate real logic.

## Fix
- Make the cast `(size_t)`, matching the sibling resolver.
- Correct the comment to the Coraza name + fix the typo.
- Remove the commented-out dead statement (keep the explanation).
- Restore the empty branch as a live compiled-out `dd()` so it is no longer an empty block.
- Drop the `#if 1`/`#endif` pair.

No behaviour change.

## Testing
Hygiene-only; the full `t/*.t` suite builds and passes unchanged. The `(size_t)` cast produces the same value for every in-range length (the string is a fixed ~29-byte date), so there is no observable output difference to assert.

## Summary by CodeRabbit

* **Refactor**
  * Removed unused debug/comment block in request body filtering without changing how the response is forwarded.
  * Simplified the pre-access handler structure by removing an unnecessary preprocessor wrapper.
  * Cleaned up header-filter comments and removed an obsolete, commented compatibility line.
* **Bug Fixes**
  * Corrected `Last-Modified` header length calculation to use the appropriate sizing type.
* **Documentation**
  * Updated header-processing commentary to better match the related response-header flow.

### DIFF
--- a/src/ngx_http_coraza_body_filter.c
+++ b/src/ngx_http_coraza_body_filter.c
@@ -495,10 +495,5 @@ ngx_http_coraza_body_filter(ngx_http_request_t *r, ngx_chain_t *in)
         return NGX_OK;
     }
 
-    if (!is_request_processed)
-    {
-        //dd("buffer was not fully loaded! ctx: %p", ctx);
-    }
-
     return ngx_http_next_body_filter(r, in);
 }
--- a/src/ngx_http_coraza_header_filter.c
+++ b/src/ngx_http_coraza_header_filter.c
@@ -297,7 +297,7 @@ ngx_http_coraza_resolv_header_last_modified(ngx_http_request_t *r, ngx_str_t nam
     p = ngx_http_time(buf, r->headers_out.last_modified_time);
 
     value.data = buf;
-    value.len = (int)(p-buf);
+    value.len = (size_t) (p - buf);
 
     return ngx_http_coraza_add_response_header(r, ctx, &name, &value);
 }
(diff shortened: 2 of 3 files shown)


ppomes:
@coderabbitai full review

eilandert:
Rebased onto main. All three hunks survived unchanged — the only conflict was positional, since #92 added the bulk response-header flush block immediately above the `msc_process_response_headers()` comment this PR was fixing. Kept the new block, applied the typo fix on top.

One thing I checked rather than assumed: removing the empty `if (!is_request_processed)` block leaves that variable still read at `ngx_http_coraza_body_filter.c:413`, so there is no unused-variable break under `-Werror`. CI on the rebased head is green across the board.

Confirmed #97 and #99 auto-merge against each other, so either order works as you said.

ppomes (approved):

PR #113, opened 2026-08-31
perf: skip inaccessible request body buffering

eilandert:
> Follow-up to [#92](https://github.com/corazawaf/coraza-nginx/pull/92). This consumes the request-side gate proposed in [corazawaf/libcoraza#118](https://github.com/corazawaf/libcoraza/issues/118) and shipped by [corazawaf/libcoraza#119](https://github.com/corazawaf/libcoraza/pull/119).

## TL;DR

When request-body inspection is disabled, the connector currently reads and copies the entire body before handing it to libcoraza, which discards it. This change lets nginx leave that body for the eventual content handler while rules scheduled after request headers are still evaluated.

In code terms, the pre-access handler now checks `coraza_is_request_body_accessible()` before requesting the body. A false result skips only nginx prereading and body submission; it still calls `coraza_process_request_body()` so phase 2 rules on non-body variables run.

**Severity:** Low (performance: avoidable request buffering, copies, and C/Go crossings when request-body access is off)

## Mechanism

- Resolve the libcoraza 1.7 request-body predicate as a required dynamic symbol. The connector already requires libcoraza 1.7, so this does not raise the dependency floor.
- Keep the existing accessible-body path unchanged, including asynchronous reads, file-backed bodies, body-limit interventions, and phase 2 processing.
- Extract the common phase 2 completion path so both accessible and inaccessible requests preserve intervention and fail-closed behavior.

## Tests

- Added an integration test that disables request-body access in phase 1, sends only one byte of a declared 64 KiB body, and requires a phase 2 header rule to respond before the rest arrives.
- The same test verifies that an inaccessible body reaches the upstream byte-for-byte and that an accessible body remains inspected.
- Added dynamic-loader contract coverage for mandatory symbol resolution and wrapper dispatch.
- Inverting the new predicate made the regression test fail its early-response and accessible-body control assertions.
- Built with nginx 1.31.3, libcoraza 1.7.0, `-Werror`, `-Wall`, and `-Wextra`.
- Full Coraza test suite: 61 files, 532 assertions passed; existing capability guards skipped HTTP/3 and one build-tree-only check.

## Performance

This removes connector-side request-body prereading and libcoraza submission when the transaction marks the body inaccessible. No throughput benchmark is included; the improvement is the removal of those operations from the disabled path.

## Security

Phase 2 is deliberately retained. Rules using request headers, URI arguments, or other non-body variables continue to run even when request-body bytes are unavailable.

## Summary by CodeRabbit

* **Bug Fixes**
  * Requests with request-body inspection disabled are no longer delayed while waiting for an inaccessible body.
  * Security rules continue evaluating headers, URI arguments, and other non-body request data.
  * Allowed request bodies continue reaching the upstream service intact.
  * Request-body inspection remains available when enabled.

* **Tests**
  * Added coverage for request-body access behavior and dynamic library integration.

### DIFF
--- a/src/ngx_http_coraza_common.h
+++ b/src/ngx_http_coraza_common.h
@@ -224,6 +224,7 @@ ngx_http_coraza_poll_after_process(ngx_http_coraza_ctx_t *ctx,
 /* ngx_http_coraza_dl.c */
 ngx_int_t ngx_http_coraza_dl_open(ngx_log_t *log);
 void ngx_http_coraza_dl_close(ngx_log_t *log);
+int ngx_http_coraza_is_request_body_accessible(coraza_transaction_t t);
 int ngx_http_coraza_is_response_body_processable(coraza_transaction_t t);
 /* Bulk header entry points (libcoraza 1.6+); guard every call with
  * ngx_http_coraza_bulk_headers_available() (forward-declared above). */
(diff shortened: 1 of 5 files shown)


ppomes (approved):

ppomes:
Thanks @eilandert. Merged!

PR #118, opened 2026-09-02
fix: reject ruleless Coraza enablement

eilandert:
## TL;DR

You can currently switch the firewall on without giving it a single rule to enforce. Nginx starts, requests flow, and the setting looks active while nothing is actually being checked. This makes that configuration fail at startup instead of quietly doing nothing.

In code terms: `ngx_http_coraza_validate_rules()` runs during postconfiguration and rejects an enabled `coraza on` when the `http` block declares no `coraza_rules` or `coraza_rules_file` anywhere.

**Severity:** `High` (`security configuration - an enabled but ruleless WAF silently allowed every request`)

## What changed

- Postconfiguration walks the merged location configurations and raises an emergency configuration error when one has `coraza on` with no rules loaded anywhere. `nginx -t` and startup both fail before workers fork.
- The check reads configuration state only. Loading `libcoraza` stays in `init_process`, after the fork.
- `t/coraza-http-level-rules.t` keeps its existing inheritance checks and adds a ruleless config that must fail with the new diagnostic.

## Scope, and what this deliberately does not catch

The validator only fires when the entire `http` block contains no rule directives. If any scope loads rules, the check stands down for the whole configuration.

That leaves a real gap: with `server A` loading rules and `server B { location / { coraza on; } }` inheriting none, B still starts and enforces nothing, because `init_process` skips building a WAF for a location whose `has_rules` is zero. Catching that needs a per-scope rule, which would reject the established layout of enabling Coraza at server level and attaching rules in child locations. That is a compatibility decision with its own regression matrix, so I left it out rather than smuggle it in here.

The added test is named for what it actually proves, the zero-rules-anywhere case, not the broader guarantee.

## Compatibility

This tightens one contract: `coraza on` is invalid when no rule directives exist anywhere in its `http` block. Server-level enable with child-location rules keeps working, as do the existing http/server/location inheritance paths.

`t/coraza-transaction-id-args.t` needed a rule added to its config to keep passing, which is a fair signal of the blast radius: anyone running `coraza on` with rules supplied through a path the module does not count will now hit a hard failure at `nginx -t`.

## Testing

`t/coraza-http-level-rules.t` covers the ruleless config: `nginx -t` must exit non-zero and print the new diagnostic, alongside the existing positive inheritance and negative-control cases. The rule added to `t/coraza-transaction-id-args.t` keeps that suite valid under the tightened contract.

I have not run the full prove suite against a built module in this session, so I am not claiming a local black-box pass; CI provides the build and runtime coverage.

## Summary by CodeRabbit

- **Bug Fixes**
  - Nginx now rejects configurations that enable Coraza without any configured rules, preventing startup with an ineffective security configuration.
  - Invalid configurations fail closed with an emergency error during validation.

- **Tests**
  - Added coverage for ruleless Coraza configurations and configuration validation failures.
  - Updated transaction ID argument coverage to include a valid Coraza rule configuration.

### DIFF
--- a/src/ngx_http_coraza_module.c
+++ b/src/ngx_http_coraza_module.c
@@ -24,6 +24,7 @@ static void *ngx_http_coraza_create_main_conf(ngx_conf_t *cf);
 static char *ngx_http_coraza_init_main_conf(ngx_conf_t *cf, void *conf);
 static void *ngx_http_coraza_create_conf(ngx_conf_t *cf);
 static char *ngx_http_coraza_merge_conf(ngx_conf_t *cf, void *parent, void *child);
+static ngx_int_t ngx_http_coraza_validate_rules(ngx_conf_t *cf);
 static ngx_int_t ngx_http_coraza_init_process(ngx_cycle_t *cycle);
 static void ngx_http_coraza_exit_process(ngx_cycle_t *cycle);
 
@@ -428,6 +429,10 @@ ngx_http_coraza_init(ngx_conf_t *cf)
 	ngx_http_core_main_conf_t *cmcf;
 	int rc = 0;
 
+	if (ngx_http_coraza_validate_rules(cf) != NGX_OK) {
+		return NGX_ERROR;
+	}
+
 	cmcf = ngx_http_conf_get_module_main_conf(cf, ngx_http_core_module);
 	if (cmcf == NULL)
 	{
@@ -476,6 +481,45 @@ ngx_http_coraza_init(ngx_conf_t *cf)
 	}
 
 	return NGX_OK;
+}
+
+static ngx_int_t
+ngx_http_coraza_validate_rules(ngx_conf_t *cf)
+{
+	ngx_http_coraza_main_conf_t *mmcf;
+	ngx_http_coraza_conf_t *lcf;
+	ngx_http_coraza_conf_t **loc_confs;
+	ngx_uint_t i;
+
+	mmcf = ngx_http_conf_get_module_main_conf(cf, ngx_http_coraza_module);
+	if (mmcf == NULL || mmcf->rules_inline != 0 || mmcf->rules_file != 0) {
+		return NGX_OK;
+	}
+
+	lcf = ngx_http_conf_get_module_loc_conf(cf, ngx_http_coraza_module);
+	if (lcf != NULL && lcf->enable == 1) {
+		goto ruleless;
+	}
+
+	if (mmcf->loc_confs == NULL) {
+		return NGX_OK;
+	}
+
+	loc_confs = mmcf->loc_confs->elts;
+	for (i = 0; i < mmcf->loc_confs->nelts; i++) {
+		lcf = loc_confs[i];
+		if (lcf->enable == 1) {
+			goto ruleless;
+		}
+	}
+
+	return NGX_OK;
+
+ruleless:
+	ngx_log_error(NGX_LOG_EMERG, cf->log, 0,
+				  "coraza: \"coraza on\" requires at least one "
+				  "inherited or configured rule");
+	return NGX_ERROR;
 }
 
 static void *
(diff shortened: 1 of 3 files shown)


ppomes:
cc @fzipi

Went looking at how the other connectors handle this before reviewing, since failing hard at startup is a behavior change:

- **ModSecurity-nginx**: no such check — `modsecurity on;` with no rules file starts and runs with nothing loaded.
- **coraza-apache**: same, and explicitly so — `child_init` builds an empty WAF with a comment saying "transactions pass through without rules".
- **coraza-caddy**: has Caddy's `Validate()` hook but it's a stub (`return nil`); an empty directives block runs ruleless.

So this makes coraza-nginx the first connector in the lineage to fail closed on an enabled-but-ruleless config, rather than matching existing behavior. For a WAF I think that's the right divergence — "looks armed, checks nothing" is exactly the misconfiguration that shouldn't survive `nginx -t` — but worth stating that it's deliberate, and it belongs in the changelog as a config-tightening note when this cuts a release.

Verified the two things the check hinges on: every srv/loc conf lands in `mmcf->loc_confs` at merge time, so no enabled scope escapes the walk; and the `coraza_rules`/`coraza_rules_file` handlers bump the main-conf counters from any scope, so rules-in-a-location-only can't false-positive. The documented gap (a server inheriting no rules while another server has some) is a fair deferral — a per-scope rule would break the enable-at-server/rules-in-locations layout.

Follow-up thought: apache and caddy have the same hole — apache would take the same shape as this (its `rules_inline`/`rules_file` counters already exist), and caddy is nearly a one-liner in that `Validate()` stub. Worth porting once this lands.

eilandert:
Both verifications match what I found, and thanks for doing the cross-connector survey — I'd only checked ModSecurity-nginx before writing this, so the apache and caddy data points are new to me.

Agreed the divergence should be stated rather than left implicit. I've rebased onto main (#119/#120 had moved under it) and rewritten the commit body to say it outright: the other three connectors accept an enabled-but-ruleless config and run it as a pass-through, this one won't, and a config that previously started will now be rejected at init. Since release-please generates the changelog from commit bodies, that puts the config-tightening note in the release entry rather than in a hand-edit that the next release would clobber.

On the port: happy to take apache after this lands, since its counters already exist and it'd be the same shape. Caddy's `Validate()` stub I'd rather leave to someone who knows whether an empty directives block is load-bearing for anyone there.

ppomes:
@fzipi when you get a sec, could you take a look at this one? It rejects enabling Coraza with no rules loaded (fail-closed rather than running an empty engine). I checked the other connectors — none of them guard this today, so it's a bit ahead of the pack and I'd like your read before we merge.

fzipi:
Took a look, and also stress-tested the "at least one rule directive" check against `SecRemoteRules`, since it's the closest thing to a rules-without-content ModSecurity flow.

Short version: it doesn't reopen the gap this PR closes, but it's a related edge case worth naming.

- `coraza_rules "SecRemoteRules https://...";` sets `has_rules`/`rules_inline` at config-parse time like any other inline rule, so it satisfies the new validator and `nginx -t` passes.
- It doesn't survive contact with libcoraza, though. Coraza's SecLang parser implements `SecRemoteRules` as a hard `errors.New("not implemented")` — it's not one of the no-op compat directives (`SecConnWriteStateLimit` and friends). So `coraza_rules_add()` fails in `init_process`, `ngx_http_coraza_build_waf` returns 0, and `init_process` returns `NGX_ERROR` — the worker fails to initialize.

So the failure mode isn't "silently armed, checks nothing" — it's "passes `nginx -t`, then every worker fails to start." Arguably worse for an operator than the case this PR fixes, since it surfaces after fork instead of at config validation.

I don't think it blocks this PR — the validator here is deliberately doing presence-of-directive checking, not full parse validation, and teaching it to distinguish "syntactically present" from "will actually load" would mean invoking the SecLang parser during config-time validation, which runs into the same dlopen-after-fork constraint this whole module is built around. Worth a follow-up issue rather than scope creep here.

Rest of the PR looks right to me — the merge-time walk and counter bump logic check out, and the documented gap (server-level enable, rules only in a sibling server) is a fair thing to defer rather than break the enable-at-server/rules-in-locations layout. LGTM.

ppomes (approved):

PR #119, opened 2026-09-02
perf: bound file-backed response inspection

eilandert:
## TL;DR

When the firewall inspects a large file being sent to a client, it used to pull the whole file into memory first, so a big download meant a matching spike in memory per request. This reads the file in fixed-size pieces instead, so inspection costs the same whether the response is 100 KB or 100 MB.

In code terms: `ngx_http_coraza_append_response_body_file()` reads file-backed response ranges through one reusable 64 KiB scratch buffer, and delayed stable file ranges keep their nginx file descriptors instead of being copied into the request pool.

**Severity:** `Low` (`performance - large inspected file responses retained body-sized request-pool allocations`)

## What changed

- File-backed ranges are validated, read in bounded chunks, and appended to Coraza until the full range is consumed.
- Short reads, invalid ranges, allocation failures, and Coraza append failures take the existing fail-closed path.
- Stable non-temporary file buffers stay file-backed while headers are delayed. Temporary upstream files keep the copy path, because nginx owns their lifetime.

## Direct I/O

The reader hands its `ngx_alloc()` scratch buffer straight to `ngx_read_file()`. Under `directio` the underlying `pread()` wants an aligned buffer and offset, so an unaligned read fails with `EINVAL`, which this path would report as `NGX_ERROR` and turn into an intervention on a perfectly clean response.

So direct I/O is toggled off around the read and restored afterwards, preserving `ngx_errno` across the restore, the same dance `ngx_output_chain_copy_buf()` does for its own unaligned reads. The runtime cases set `directio 512` so the clean and blocked file paths both exercise that branch.

## A note on the delayed-header cap

Retaining file buffers changes what `pending_bytes` accounts for: a file-backed range now charges its logical length against the cap while costing no memory. Medium-sized static responses therefore reach the early-header-flush threshold sooner than they did when only in-memory copies counted. The behaviour is intentional, but the 256 KiB body in `t/coraza-response-body.t` sits under the cap and does not exercise it.

`ngx_read_file()` also advances `file->offset` as a side effect. That field is dead state for this buffer: `ngx_output_chain.c` never reads it and the sendfile path works from `buf->file_pos`, so downstream delivery is unaffected.

## Testing

`t/coraza-response-body-file-reader-contract.t` covers the bounded reader's contract checks. The runtime cases in `t/coraza-response-body.t` put the blocking marker past the first 64 KiB chunk and exercise the clean and blocked paths.

I did not run the full prove suite locally in this session; the deterministic lint roster reported no confirmed defect in the changed lines, and clang-tidy could not parse without the generated nginx build headers, so CI provides the compile and runtime coverage.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved inspection of large file-backed response bodies by processing them in bounded chunks, reducing memory usage.
  * Preserved fail-closed behavior when response-body inspection encounters read failures.
  * Improved delayed response handling for file-backed buffers.

* **Tests**
  * Added coverage for large response bodies spanning multiple buffers.
  * Added validation for blocking content detected beyond the first file chunk.
  * Added checks for bounded file reading, error handling, and final-buffer inspection.

### DIFF
--- a/src/ngx_http_coraza_body_filter.c
+++ b/src/ngx_http_coraza_body_filter.c
@@ -12,6 +12,8 @@
 
 static ngx_http_output_body_filter_pt ngx_http_next_body_filter;
 
+#define NGX_HTTP_CORAZA_RESPONSE_BODY_FILE_CHUNK_SIZE (64 * 1024)
+
 /*
  * Length of the body data ngx_http_coraza_read_body_data would return,
  * without performing the read/allocation -- lets callers reject an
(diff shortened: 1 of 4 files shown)


ppomes:
@coderabbitai full review

ppomes (approved):

PR #120, opened 2026-09-02
fix: reject premature response file EOF

eilandert:
## TL;DR

If a file being sent to a client ends earlier than it claimed, the firewall was treating that short read as a normal finished response. The bytes that never arrived were never inspected, so a truncated file could slip past the checks. This rejects the response instead.

In code terms: `ngx_http_coraza_read_body_data()` returns `NGX_ERROR` when `ngx_read_file()` hits EOF before `file_last`, replacing a bare `break` that handed Coraza a short body and let it evaluate phase 4 against fewer bytes than nginx would stream.

**Severity:** `Low` (`correctness - truncated file-backed responses were accepted as complete`)

## Why fail closed

The sibling reader, `ngx_http_coraza_append_response_body_file()`, already returns `NGX_ERROR` with a near-identical message on the same condition. The two file readers now agree, which is the point: a partial read is indistinguishable from a complete one downstream, so the only safe reading of a short EOF is that inspection is incomplete.

The check cannot fire on a legitimate complete read. `n == 0` is only reachable inside `while (remaining > 0)`, and `remaining` hits zero exactly when the advertised range has been consumed, so a full read leaves the loop before the test. A genuinely empty range returns earlier still.

## Testing

`t/coraza-response-file-eof.t` compiles the production reader against the nginx headers and asserts both directions: a premature EOF fails closed, and a complete 18-byte read returns `NGX_OK` with the expected length, so an over-eager fix would be caught too.

I verified it is not vacuous. Reverting the production hunk and re-running turns assertion 2 red (`got: '256'`, the harness's return from the `!= NGX_ERROR` branch); restoring it passes all three.

The harness resolves the module source by walking up for `src/ngx_http_coraza_common.h`, matching `find_root()` in `coraza-h3-protocol-map.t`, rather than assuming a fixed relative path. It skips rather than misreports when the module tree or `coraza/coraza.h` is absent, and its guard checks every include directory the compile needs, so a partial `TEST_NGINX_SOURCE` tree skips instead of failing. I ran it both from `t/` and from a staged copy inside an unpacked `nginx-tests-*` directory, the layout CI proves from.

## Summary by CodeRabbit

* **Bug Fixes**
  * Response processing now reports an error when a file-backed response ends unexpectedly, preventing incomplete data from being treated as valid.

* **Tests**
  * Added coverage verifying correct handling of both premature file termination and complete response data.

### DIFF
--- a/src/ngx_http_coraza_body_filter.c
+++ b/src/ngx_http_coraza_body_filter.c
@@ -85,7 +85,10 @@ ngx_http_coraza_read_body_data(ngx_http_request_t *r, ngx_buf_t *buf,
             }
 
             if (n == 0) {
-                break;
+                ngx_log_error(NGX_LOG_ERR, r->connection->log, 0,
+                    "coraza: response body file ended at %O of %O bytes",
+                    offset, buf->file_last);
+                return NGX_ERROR;
             }
 
             total_read += (size_t) n;
--- a/t/coraza-response-file-eof.t
+++ b/t/coraza-response-file-eof.t
@@ -1,0 +1,181 @@
+#!/usr/bin/perl
+
+# Executable unit test for file-backed response reads.  The production source
+# is included directly; unused filter sections are discarded by the linker.
+
+use warnings;
+use strict;
+
+use Test::More;
+use File::Temp qw(tempdir);
+use FindBin;
+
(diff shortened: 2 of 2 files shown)


ppomes:
@coderabbitai full review

ppomes (approved):

PR #124, opened 2026-09-03
fix: intervention poll consistency in the pre-access handler

eilandert:
The two in-loop intervention polls in the pre-access handler — one per chunk in the file-buffered body path, one per buffer in the in-memory path — had no `r->error_page` check. The final poll in the same function, and the sites in `rewrite.c`, `header_filter.c` and the body filter, all yield once nginx is already streaming the configured error page, so a fresh intervention raised on the error-page pass cannot finalize the request a second time. Both in-loop sites now yield the same way.

The file-path guard yields explicitly with `NGX_DECLINED` rather than jumping to the cleanup label with `rc` still `NGX_OK` — that shape would have reported the whole body inspected while the remaining chunks were never appended. The trailing `intervention_triggered` arming is exempted for that value, since none of the sibling sites arm the flag on a deliberate yield.

On the error-page pass the final poll already returned before `ngx_http_coraza_process_request_body_phase()`, so phase 2 never ran there. The earlier yields stop appending chunks that would not have been processed; they do not remove any inspection that used to happen.

## Three unrelated nits in the same area

- `coraza_rules` and `coraza_rules_file` declared `offset = offsetof(..., enable)`, which their handlers never read. Corrected to `0`, matching `coraza_transaction_id`.
- Removed the `lcf->rules == mmcf->rules` dedup branch in `init_process`. `mmcf->rules` is never populated, so it could not match; the real per-loc-conf dedup loop below it is unchanged.
- The log handler returned `NGX_ERROR` on a NULL ctx. LOG-phase handlers return `NGX_OK` by nginx contract.

## Testing

`t/coraza-request-body-error-page-reentry.t` routes a backend 404 into a `coraza on` named location whose phase-2 `REQUEST_BODY` rule would fire if the second pass re-inspected the body. Three body sizes: one that fits in the header preread (single buffer, `last_buf`, the in-memory loop breaks before its poll), a ~4 KiB one larger than what remains of the preread buffer but under `client_body_buffer_size` so nginx reads the rest into a second in-memory buffer and the in-memory in-loop poll runs, and one larger than `client_body_buffer_size` and the 64 KiB file chunk so the file loop's in-loop poll site actually iterates. `$request_body_file` in the access log pins the memory-versus-file split, so the two re-entry cases cannot silently exercise the same path. Direct POST controls prove the rule and all three body paths work without `error_page`. The test pins the observable contract — 404 served, no second finalize, worker healthy — not the internal poll ordering.

The in-loop yields are reached but not status-observable here, and the file header says so: phase-2 rules only run in the final process-request-body step, so the in-loop polls return 0 on the re-entry pass, and a deny on the first pass returns from the handler before the loop. Deleting the in-memory in-loop guard leaves the file green; the final poll's yield is what the 404 assertions pin.

Rebased onto #134, which renamed the file reader's cleanup label; the file-path yield jumps to `done` now.

Full suite against nginx 1.31.4 with libcoraza 1.7: 429 tests, all passing. Module builds clean under the CI `-Werror` flags.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed request-body inspection when requests re-enter through an `error_page` route.
  - Prevented previously handled requests from being finalized a second time, avoiding incorrect status changes such as turning a 404 into a 403.
  - Improved handling for both in-memory and file-buffered request bodies.
  - Ensured worker processes remain stable during these error-page flows.

- **Tests**
  - Added coverage for multi-chunk request bodies, error-page routing, and worker stability.

### DIFF
--- a/src/ngx_http_coraza_log.c
+++ b/src/ngx_http_coraza_log.c
@@ -27,7 +27,8 @@ ngx_http_coraza_log_handler(ngx_http_request_t *r)
     ctx = ngx_http_get_module_ctx(r, ngx_http_coraza_module);
 
     if (ctx == NULL) {
-        return NGX_ERROR;
+        /* LOG-phase handlers must return NGX_OK */
+        return NGX_OK;
     }
 
     if (ctx->logged) {
--- a/src/ngx_http_coraza_module.c
+++ b/src/ngx_http_coraza_module.c
@@ -369,13 +369,13 @@ static ngx_command_t ngx_http_coraza_commands[] = {
 	 NGX_HTTP_LOC_CONF | NGX_HTTP_SRV_CONF | NGX_HTTP_MAIN_CONF | NGX_CONF_TAKE1,
 	 ngx_conf_set_rules,
 	 NGX_HTTP_LOC_CONF_OFFSET,
-	 offsetof(ngx_http_coraza_conf_t, enable),
+	 0,
 	 NULL},
 	{ngx_string("coraza_rules_file"),
 	 NGX_HTTP_LOC_CONF | NGX_HTTP_SRV_CONF | NGX_HTTP_MAIN_CONF | NGX_CONF_TAKE1,
 	 ngx_conf_set_rules_file,
 	 NGX_HTTP_LOC_CONF_OFFSET,
-	 offsetof(ngx_http_coraza_conf_t, enable),
+	 0,
 	 NULL},
 	{ngx_string("coraza_transaction_id"),
 	 NGX_HTTP_LOC_CONF | NGX_HTTP_SRV_CONF | NGX_HTTP_MAIN_CONF | NGX_CONF_TAKE1,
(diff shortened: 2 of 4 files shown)


eilandert:
Checked this against `main` before deciding anything. The guard is the same one the final poll in this function has carried since before this change (`pre_access.c:389` on `main`), alongside `rewrite.c:315` and `header_filter.c:616`; this PR adds it to the two in-loop sibling sites that lacked it. `ctx->intervention_triggered` already is the dedicated state for Coraza-originated denials — the handler returns on it before any body work. `r->error_page` covers the other case, a non-Coraza error routed through `error_page`, which no Coraza flag can know about.

On "does not inspect the remaining body": the final poll on `main` returns `NGX_DECLINED` on `r->error_page` before `ngx_http_coraza_process_request_body_phase()`, so phase 2 never ran on an error-page pass. The earlier yields stop appending chunks that were never going to be processed. Nothing that used to be inspected is skipped now.

The coverage point stands — all three error_page tests were GET-only. `t/coraza-request-body-error-page-reentry.t` carries a body through a backend 404 into a coraza-on named location with a phase-2 body rule, with both an in-memory body and one large enough to make the file loop iterate past its first chunk. Direct POST controls prove the rule fires without error_page in the way; the first draft passed on nothing (`return` answers before pre-access, static answers POST with 405, REQUEST_BODY is empty without a body processor) and the controls caught it.

ppomes:
@coderabbitai full review

ppomes:
@coderabbitai full review

ppomes (approved):

PR #126, opened 2026-09-03
fix: submit the request body to Coraza exactly once

eilandert:
The pre-access handler walked `r->request_body->bufs` and appended the body to the transaction every time it ran. nginx re-runs the PREACCESS phase on the same request in a few situations — an `error_page` internal redirect, a named-location jump, another PREACCESS module returning `NGX_DONE` and resuming — and each re-run submitted the whole body again. The engine then held the body twice: `REQUEST_BODY` duplicated, `ARGS_POST` values doubled, `SecRequestBodyLimit` tripping at half its configured size, phase 2 evaluated twice.

`ctx->body_submitted` is set before the body is appended and processed, and a later re-entry short-circuits with `NGX_DECLINED`. The submit block also guards `r->request_body == NULL` at the top, and the temp-file/in-memory-chain invariant is now written down where the walk happens: `single_buf` and `clean_file` are always set before the read, so a body spilled to a temp file has no live in-memory remainder to also walk.

## Testing

`t/coraza-request-body-single-submit.t` pins the single-submission contract directly. A 12-byte body against `SecRequestBodyLimit 20` with `SecRequestBodyLimitAction Reject` passes: a doubled submission would be 24 bytes and answer 413. The limit is exclusive at the boundary, which `t/coraza-request-body.t`'s `/bodylimitreject` pair already establishes, so the margin is deliberate.

On why the test does not construct a second PREACCESS pass: nothing in this suite can. The only modules exercised anywhere under `t/` are `http`, `proxy`, `auth_request`, `http_v2`, `http_v3`, `rewrite`, `sub` and `unix`, and none of them re-enters PREACCESS after the body is read. `auth_request` runs in ACCESS. The oracle is the observable limit behaviour, which is what the double submission actually breaks in production.

An `ARGS_POST` duplication assertion was in an earlier revision and dropped: it never matched the rule in CI and the cause was not established, so it asserted nothing. The suite's only other `ARGS_POST` case runs with `SecRequestBodyAccess Off`, so it does not demonstrate matching either. Rather than ship a vacuous check it is left out.

Full suite against nginx 1.31.4 with libcoraza 1.7: 421 tests, all passing. Module builds clean under the CI `-Werror` flags.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented request bodies from being inspected or submitted more than once during request processing.
  * Improved handling of buffered request bodies to avoid duplicate processing.
  * Requests without an available body now exit safely without attempting submission.

* **Tests**
  * Added coverage verifying that each request body is submitted to the inspection engine exactly once.

### DIFF
--- a/src/ngx_http_coraza_common.h
+++ b/src/ngx_http_coraza_common.h
@@ -87,6 +87,7 @@ typedef struct {
 
     unsigned waiting_more_body:1;
     unsigned body_requested:1;
+    unsigned body_submitted:1;
     unsigned processed:1;
     unsigned logged:1;
     unsigned intervention_triggered:1;
--- a/src/ngx_http_coraza_pre_access.c
+++ b/src/ngx_http_coraza_pre_access.c
@@ -295,13 +295,33 @@ ngx_http_coraza_pre_access_handler(ngx_http_request_t *r)
         }
     }
 
-    if (ctx->waiting_more_body == 0)
+    if (ctx->waiting_more_body == 0 && ctx->body_submitted == 0)
     {
         int ret = 0;
         int already_inspected = 0;
         ngx_int_t rc;
 
         dd("request body phase is ready to be processed");
+
+        if (r->request_body == NULL) {
+            /*
+             * Nothing to submit (e.g. body access predicate flipped between
+             * checks, or nginx never allocated a request_body). Nothing was
+             * sent to Coraza, so there is no submission to mark: leave
+             * body_submitted clear and let the phase continue normally.
+             */
+            return NGX_DECLINED;
+        }
+
+        /*
+         * Mark the body as submitted before doing the append/process work so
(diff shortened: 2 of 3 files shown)


ppomes:
@coderabbitai full review

eilandert:
On the PREACCESS re-entry nitpick: same ask CodeRabbit raised and withdrew on #121 when this file moved here. No module in the CI nginx besides coraza hooks PREACCESS, so the test cannot drive a second invocation of the handler; that needs a second PREACCESS module built into the CI image, a CI change rather than a test edit and out of scope here. The file header says so. What the test pins instead is the single-invocation contract: `SecRequestBodyLimit 20` against a 12-byte body, where a doubled submission would be 24 bytes and reject with 413.

ppomes (approved):

PR #127, opened 2026-09-03
test(h2): run the phase-4 blocking assertions in CI; document coraza_delay_response_headers

eilandert:
The four phase-4 assertions in `t/coraza-h2.t` and `t/coraza-proxy-h2.t` were gated behind `$ENV{TEST_NGINX_UNSAFE}`, which nothing in CI sets, so response-body blocking over HTTP/2 has never actually been exercised there.

The eight `SKIP` blocks (four per file) are un-gated, and each `is($frame, undef, ...)` is replaced with a positive pair: the expected `:status` on the HEADERS frame, plus an `unlike()` proving the blocked upstream body text never appears in a DATA frame. Absence of a DATA frame also follows from a reset connection, a GOAWAY, a timeout or an empty 200, so the old assertion could not tell a correct block from a broken response; `:status` plus a body-leak check can. Plans bumped to match: 20 → 24 and 23 → 27.

The proxy-h2 "nothing to detect" negative control already ran through the `HTTP2`-based loop at the end of `t/coraza-proxy-h2.t`; only its superseded `http_get()` predecessor is commented out, so nothing needed restoring there.

## README

`coraza_delay_response_headers` gets its own directive section — syntax, context (`http`, `server`, `location`), default (`on`), and what turning it off costs: a late phase-4 intervention can no longer replace a response whose headers have already gone out. The directive count goes from four to five.

A short source-build note: `./configure` needs `--with-cc-opt="-Wno-unused-function"` because the installed `coraza/coraza.h` carries libcoraza 1.7.0's cgo preamble, whose `static` callback shims are unused in any given translation unit, and nginx builds with `-Werror`. The Debian package build sets its own CFLAGS and does not need it.

## Testing

Built with `--with-http_v2_module` against nginx 1.31.4 and libcoraza 1.7. Both changed files pass with their new assertions live (24 and 27). Full suite: 477 tests, one failure in `t/coraza-h2-forbidden-headers.t` ("h2: no synthetic Connection" gets no response at all) that I could not attribute to this change — it is untouched here, the module source on this branch is identical to `main`, and upstream's build-test workflow is green on `57ac5da` with the same pins, so I am treating it as local to my h2 build rather than hiding it.

## Summary by CodeRabbit

- **Documentation**
  - Added source-build guidance for avoiding compiler warnings treated as errors.
  - Documented the `coraza_delay_response_headers` directive and updated the directive count.

- **Tests**
  - Expanded HTTP/2 coverage for redirect and blocking responses.
  - These scenarios now run consistently and verify both response status codes and that blocked content is not returned.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,6 +56,17 @@ Or, to build a dynamic module:
 Note that when building a dynamic module, your nginx source version
 needs to match the version of nginx you're compiling this for.
 
+When building from source, pass `-Wno-unused-function`. The installed
+`coraza/coraza.h` carries libcoraza's cgo preamble, whose `static` callback
+shims are unused in any given translation unit, and nginx builds with `-Werror`:
+
+```
+./configure --add-dynamic-module=/path/to/coraza-nginx --with-compat \
+    --with-cc-opt="-Wno-unused-function"
+```
+
+The Debian package build sets its own CFLAGS and does not need this flag.
+
 Further information about nginx third-party add-ons support are available here:
 http://wiki.nginx.org/3rdPartyModules
 
@@ -83,7 +94,7 @@ sets `-ldl` for you on Linux/macOS.
 # Usage
 
 coraza for nginx extends your nginx configuration directives.
-It adds four new directives:
+It adds five new directives:
 
 coraza
 ------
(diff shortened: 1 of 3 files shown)


ppomes:
@coderabbitai full review

ppomes (approved):

PR #129, opened 2026-09-03
test: delayed-header buffer-flag matrix and slow-client coverage

eilandert:
## TL;DR

Expand delayed-response coverage across nginx buffer flags, large bodies, slow clients, and downstream disconnects. Explicit holding and release markers prove cancellation occurs while the upstream response is still incomplete.

The matrix covers memory-backed and recycled buffers, flush/sync flags, split headers and bodies, and a response larger than 1 MiB. It deliberately does not claim temp-file spill coverage.

**Severity:** None — test coverage only.

The tests enable response-body access where the phase-4 oracle needs it and make the slow-client daemon fail loudly if the client-disconnect release handshake is absent.

## Summary by CodeRabbit

* **Tests**
  * Added coverage for delayed response buffering with repeated in-memory buffer recycling.
  * Added validation for responses larger than the delayed-body limit, including early header flushing and continued streaming.
  * Added coverage for client disconnects during transfers, verifying subsequent requests remain healthy.
  * Added byte-for-byte response validation and logging checks for early header flushing.

### DIFF
--- a/t/coraza-delayed-buffer-flags.t
+++ b/t/coraza-delayed-buffer-flags.t
@@ -1,0 +1,402 @@
+#!/usr/bin/perl
+
+# Tests for Coraza-nginx connector (delayed-header buffer chains).
+#
+# t/coraza-delayed-file-buffer.t covers the uninspected file-backed clone
+# branch (large file, sendfile off, phase-4 pass). t/coraza-response-body-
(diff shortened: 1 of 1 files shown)


ppomes:
@coderabbitai full review

ppomes (approved):

PR #130, opened 2026-09-03
test: prove ARGS_POST matches and move source checks to lint

eilandert:
## TL;DR

Add a positive `ARGS_POST` assertion and move source-structure checks out of the runtime suite into `lint/`. This keeps nginx integration tests behavioral while preserving the connector contract checks as fast deterministic lint tests.

Six source-grep tests move to `lint/`, including the response-body file-reader contract. The request-body cleanup check now requires the intervention result to be assigned and both return paths to reach the cleanup label, and the upgrade test documents the remaining drain limitation.

Rebased onto #134 and #136, which edited two of the moved files in place (per-worker static chunk buffer, `done` cleanup label); those edits are carried in the `lint/` copies. The response-body file-reader contract also requires the full `in_file && file` predicate before the bounded reader call, and accepts the failure site in both its inline and #122 helper form; that file's text is identical to the copy #122 edits under `t/`, so the two land cleanly in either order.

**Severity:** None — tests and test organization only.

`prove lint/*.t` passes all 42 assertions, against `main` and against `main` with #122 merged.

## Summary by CodeRabbit

- **Tests**
  - Added coverage for URL-encoded POST argument matching and upgraded-connection handling.
  - Expanded regression checks for request and response body processing, header forwarding, error propagation, and dynamic library integration.
  - Strengthened validation of request-body processing behavior and response-body file inspection.

- **Documentation**
  - Added guidance for running and understanding source-text lint checks.

- **Chores**
  - Integrated source-text lint checks into the full test workflow for earlier detection of regressions.


ppomes:
@coderabbitai full review

eilandert:
@coderabbitai full review

eilandert:
@coderabbitai full review

ppomes:
Six of the fixes just landed on `main` (#118, #122–#126). GitHub reports this one as conflicting because `t/coraza-response-body-file-reader-contract.t` was modified on `main` (#122) while this branch moves and rewrites it under `lint/`. git merges it automatically (it folds the #122 changes into the moved file), so bringing `main` into the branch (a merge, so the history stays as is) should be painless — but please give the merged `lint/coraza-response-body-file-reader-contract.t` a look afterwards, since the automatic fold is a guess, not a review. CI and CodeRabbit will re-run on the update and we will merge from there — thanks, and thanks for working through all the review threads.

eilandert:
Merged `main` in (merge, not rebase) — now mergeable at `0b3ef39`.

Git folded #122's changes into the copy this branch moved to `lint/`, with no conflict. Reviewed rather than taken on trust, as you asked:

- all three of #122's additions are present in the moved file — the tightened `in_file`/`file` guard regex, the widened fail-closed alternation, and the new delayed-buffer block asserting five internal-error helper calls and no bare `NGX_ERROR` return;
- no stale copy survives at the old `t/` path;
- the file declares `done_testing()`, so there is no plan count to desync, and `$root = "$FindBin::Bin/.."` still resolves from `lint/`.

Verified: the moved file passes 11/11 and the whole `lint/` suite passes 45/45. The folded assertions are not vacuous — relaxing the `in_file`/`file` guard in `ngx_http_coraza_body_filter.c` fails assertion 5 and only assertion 5.

ppomes (approved):

PR #134, opened 2026-09-04
perf: trim request and connection-header hot paths

eilandert:
## TL;DR

Large request bodies and common connection-header handling do small amounts of avoidable work on every request or response. This sweep reuses scratch storage and carries already-known address and header values forward without changing inspection behavior.

**Severity:** Low (performance and cleanup; request behavior is unchanged)

## Request body

- The file-backed reader uses one worker-lifetime 64 KiB buffer instead of allocating and freeing scratch space per request. The reader is synchronous on nginx's single-threaded worker loop, so requests cannot overlap inside it.
- Empty in-memory buffers no longer cross into libcoraza. End-of-chain handling still runs for every buffer, including an empty final buffer.
- `request_body_in_single_buf` is no longer set. On supported nginx versions it does not control the memory/file split or body-chain shape.
- The reader retains its private file descriptor because `ngx_read_file()` advances `ngx_file_t.offset` bookkeeping even when the platform uses `pread()`.

## Connection metadata

- The rewrite handler formats the local address already cached by its first `ngx_connection_local_sockaddr()` call instead of re-entering the helper.
- The header filter uses `sizeof(literal) - 1` for Connection values and the end pointer returned by `ngx_sprintf()` for Keep-Alive, avoiding follow-up string scans.
- Unix-domain handling and the HTTP/2 and HTTP/3 early-return behavior are unchanged.

### DIFF
--- a/src/ngx_http_coraza_header_filter.c
+++ b/src/ngx_http_coraza_header_filter.c
@@ -309,6 +309,7 @@ ngx_http_coraza_resolv_header_connection(ngx_http_request_t *r, ngx_str_t name, 
     ngx_http_coraza_ctx_t *ctx = NULL;
     ngx_http_core_loc_conf_t *clcf = NULL;
     char *connection = NULL;
+    size_t connection_len = 0;
     ngx_str_t value;
 
     clcf = ngx_http_get_module_loc_conf(r, ngx_http_core_module);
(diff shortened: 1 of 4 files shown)


ppomes (approved):

ppomes:
@fzipi one thing worth a look before this goes in: it drops `r->request_body_in_single_buf = 1` from the pre-access handler. ModSecurity-nginx still sets that flag (with a TODO admitting it is not sure what it does), so removing it here means we no longer match upstream on that point. The reasoning in the PR is that the flag does not control the memory/file split or the body-chain shape on the nginx versions we support, which matches what I see — but since it is a behavioural knob inherited from ModSec I would rather you sign off on dropping it. The rest is straightforward hot-path trimming (buffer reuse for file-backed bodies, sizeof/pointer lengths instead of strlen) with no behaviour change.

PR #135, opened 2026-09-04
perf: deduplicate WAFs by ordered rule content

eilandert:
## TL;DR

Several server blocks that load identical firewall rules currently compile and store separate copies. This change shares one compiled copy when the ordered rule text is identical, while different or differently ordered policies remain separate.

**Severity:** Low (startup time and per-worker memory use)

## Mechanism

`ngx_http_coraza_init_process()` first checks rules-array pointer identity, then compares the ordered `(type, value)` entries. Rule order remains part of the key because SecLang evaluates rules sequentially.

The comparison uses the stored directive text. Both rule setters preserve the raw token, and `ngx_http_coraza_build_waf()` passes that same text to libcoraza from a single worker working directory. If path resolution becomes context-dependent, the comparison must be tightened or removed.

## Ownership

`waf_owner` marks the location configuration that built each handle. Borrowers keep the owner bit clear, and `exit_process()` frees only owners, so every shared handle has one release. The main configuration retains ownership of its fallback handle.

`t/coraza-waf-dedup.t` covers identical policies, distinct policies, and the same two stateful rules in opposite order. The order pair has different request outcomes, so both the policy result and the WAF initialization count detect an incorrect merge.

## Summary by CodeRabbit

* **Performance**
  * Improved resource usage by reusing Web Application Firewall instances when configurations contain identical rules.

* **Bug Fixes**
  * Ensured configurations with different rules, or the same rules in a different order, remain separately evaluated.
  * Improved WAF lifecycle handling to prevent duplicate cleanup and preserve reliable worker shutdown behavior.

### DIFF
--- a/src/ngx_http_coraza_common.h
+++ b/src/ngx_http_coraza_common.h
@@ -111,6 +111,17 @@ typedef struct {
     coraza_waf_t               waf;
     ngx_array_t               *rules;       /* of ngx_http_coraza_rule_entry_t */
 
+    /*
+     * WAF ownership.  init_process dedups WAFs across loc_confs whose rules
+     * arrays are content-identical, so several loc_confs can point at the
+     * same coraza_waf_t.  Exactly one of them -- the one that actually built
+     * it -- has waf_owner set, and only that one frees the handle in
+     * exit_process.  Sharers leave waf_owner clear and must never free.
+     * A loc_conf that falls back to mmcf->waf is never an owner either;
+     * mmcf owns that handle.
+     */
+    ngx_flag_t                 waf_owner;
+
     ngx_flag_t                 enable;
     ngx_flag_t                 has_rules;
     ngx_flag_t                 delay_response_headers;
--- a/src/ngx_http_coraza_module.c
+++ b/src/ngx_http_coraza_module.c
@@ -562,6 +562,7 @@ ngx_http_coraza_create_conf(ngx_conf_t *cf)
 	 *     conf->transaction_id = NULL;
 	 *     conf->has_rules = 0;
 	 *     conf->delay_response_headers = 0;
+	 *     conf->waf_owner = 0;
 	 */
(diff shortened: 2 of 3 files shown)


ppomes:
@coderabbitai full review

ppomes:
@coderabbitai full review

eilandert:
Fixed the documentation-coverage warning in `fa611a0`. `src/ngx_http_coraza_module.c` now documents the zero-initialized ownership contract in `ngx_http_coraza_create_conf()` and the failure contract of `ngx_http_coraza_build_waf()`.

ppomes:
Six of the fixes just landed on `main` (#118, #122–#126), which puts this one in conflict on `src/ngx_http_coraza_module.c` (against #124). Could you bring `main` into the branch (a merge, so the history stays as is) when you have a minute? CI and CodeRabbit will re-run on the update and we will merge from there — thanks, and thanks for working through all the review threads.

eilandert:
Merged `main` in (merge, not rebase) — now mergeable at `7390570`.

The conflict in `src/ngx_http_coraza_module.c` was confined to the comment above the dedup loop: #124 reworded it while this branch replaced the loop with the content-comparison version and added the main-WAF fast path ahead of it. Kept this branch's code and its comment, which documents what #124's one-liner described plus the ownership rule the loop now depends on.

The fast path leaves `waf_owner` clear, which is correct because `create_loc_conf()` zeroes it, so a location sharing the main WAF never frees a handle it does not own.

Verified in a throwaway container against nginx 1.31.4 and libcoraza v1.7.0: the module compiles under the CI `-Werror` flags, and `t/coraza-waf-dedup.t` passes 14/14 — including the negative controls that two blocks with different rules keep separate WAFs, and the count assertion that six rules arrays still collapse to five.

ppomes (approved):

PR #136, opened 2026-09-04
fix: use libcoraza's interruption result and required bulk APIs

eilandert:
## TL;DR

The connector carried a hardcoded copy of the firewall engine's "interrupted" result even though the engine already defines it. The values match today, but a future library change could make blocked requests pass without a diagnostic.

**Severity:** Medium (future fail-open risk at the rule-phase boundary)

## Result handling

`CORAZA_INTERRUPTION` is an enum value in `coraza.h`, not a preprocessor macro. The local `#ifndef` therefore fired on every build and replaced later uses with the literal `1`. Removing it binds comparisons to the value declared by the installed header.

`ngx_http_coraza_poll_after_process()` also treats a negative phase result as an internal error. Without that branch, a failed phase has no intervention to fetch and would continue uninspected. Current libcoraza 1.7 header-phase functions return only success or interruption, so this is a forward-compatible failure mode rather than a currently reachable one.

## Dead capability paths

libcoraza 1.7 or newer is enforced at configure time and again when each worker loads the library. Both bulk-header functions are required symbols. The optional-symbol checks and their unreachable per-header branches are removed; allocation, packing, and submission failures still use the existing per-header fallback.

The dynamic-library contract checks pin both bulk-header functions to required symbol loading, and the intervention-poll fixture now describes the enforced return-value contract.

## Summary by CodeRabbit

* **Compatibility**
  * Support now requires libcoraza 1.7.0 or later.
  * Older libcoraza versions no longer load successfully.

* **Performance**
  * Request and response headers use bulk submission by default, with per-header processing retained when packing or submission fails.

* **Reliability**
  * Processing errors now fail closed with an internal server error.
  * Intervention handling uses the standardized libcoraza status results.

### DIFF
--- a/src/ngx_http_coraza_common.h
+++ b/src/ngx_http_coraza_common.h
@@ -82,7 +82,8 @@ typedef struct {
     /* When non-NULL, ngx_http_coraza_add_response_header() collects pairs
      * here instead of making one cgo call each, so the header filter can
      * submit the whole set via coraza_add_response_headers() in a single
-     * crossing (libcoraza 1.6+).  NULL => direct per-header cgo path. */
+     * crossing.  It is NULL outside collection, during fallback replay, or
+     * when ngx_array_create() cannot arm the collector. */
     ngx_array_t *resp_hdr_collect;
 
     unsigned waiting_more_body:1;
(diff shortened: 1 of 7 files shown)


ppomes:
@coderabbitai full review

ppomes (approved):

PR #137, opened 2026-09-04
test: cover the bounded chunked response-file reader

eilandert:
## TL;DR

When a response is big enough that nginx spills it to disk, we read it back in
64 KiB pieces and pass each piece to the rules engine for inspection. That
reader arrived in #119 with no test that actually runs it, so a bug in how it
walks through the file would have gone unnoticed. This adds one, built so that
breaking the reader in any of ten plausible ways makes it fail.

Test-only: no `src/` change.

**Severity:** none — this closes a coverage gap, it does not fix a defect.

## What was uncovered

`t/coraza-response-file-eof.t` covered `ngx_http_coraza_read_body_data()`'s
`n == 0` fail-closed branch. The bounded chunked reader #119 added,
`ngx_http_coraza_append_response_body_file()`, has the same
premature-EOF branch and had no executable test at all. Both are correct by
inspection; only one was verified by running it.

The fixture is not Test::Nginx. It compiles `ngx_http_coraza_body_filter.c`
directly with `#define static` and stubs `ngx_read_file()`, which is what makes
these branches drivable — the temp-file failure paths are unreachable through
the Perl harness, since nginx names a fresh temp file per request and core's own
`ngx_create_temp_file` fails first.

## Scenarios

Truncating the backing file mid-range so the second read returns 0; the intact
multi-chunk control; a chunk Coraza rejects; short reads; a mid-file range; an
empty range; a read error; an inverted range; and scratch-buffer allocation
failure. Each returns a distinct exit code, mapped to a description so a
failure names its own cause.

## Evidence

Both suites green: this file 3/3, `coraza-response-body-file-reader-contract.t`
8/8, against nginx 1.31.4.

Every scenario is mutation-checked — each mutation below turns exactly one
scenario red, and no other:

| Mutation to `ngx_http_coraza_body_filter.c` | Scenario |
| --- | --- |
| chunked reader's `n == 0` no longer fails closed | 3 |
| whole-range reader's `n == 0` no longer fails closed | 1 |
| ignore `coraza_append_response_body()`'s rejection | 8 |
| clamp always reads a full chunk | 7 |
| `offset += size` instead of `offset += n` | 11 |
| append `(int) size` instead of `(int) n` | 11 |
| start at `0` instead of `buf->file_pos` | 13 |
| report success after `ngx_read_file()` returns `NGX_ERROR` | 15 |
| accept `file_last < file_pos` | 16 |
| report success when scratch-buffer allocation fails | 17 |

The short-read, size-clamp, and starting-offset mutations matter most, because
they are the ones a reader-walk bug actually looks like. `offset += size` is the sharpest: with a stub that never short-reads
it is indistinguishable from correct, but in production it advances past bytes
`ngx_read_file()` never returned, so the remainder of the range is never
inspected. Covering it needed a stub that returns short reads and a range whose
length is not a whole multiple of the chunk size — with an aligned range, every
read is a full chunk and the clamp's tail branch never runs.

The direct-I/O stub guard also has a portability control: compiling with
`NGX_HAVE_F_NOCACHE=1` and `NGX_HAVE_O_DIRECT=0` passes with the guard and fails
at macro expansion when the guard is removed.

## Not covered

The `NGX_HAVE_ALIGNED_DIRECTIO` block: the fixture leaves `buf->file->directio`
unset, so `ngx_directio_off`/`_on` exist only to satisfy the linker. The
contract test asserts that block's shape by regex over the source, which catches
deletion but not a behavioural change. Worth a follow-up, not this PR.

## Summary by CodeRabbit

- **Tests**
  - Expanded response-file coverage for chunked reads, premature end-of-file conditions, empty and inverted ranges, nonzero offsets, short reads, rejected chunks, read errors, truncation, and allocation failures.
  - Added validation for intervention signaling, failure finalization, successful controls, and preservation of complete multi-chunk responses.
  - Improved failure diagnostics with scenario-specific details and process termination reporting.
  - Enhanced test portability by supporting standard system locations for required headers.

### DIFF
--- a/t/coraza-response-file-eof.t
+++ b/t/coraza-response-file-eof.t
@@ -55,8 +55,13 @@ for my $dir (@needed) {
 
 plan skip_all => 'module source tree not found'
     unless defined $root;
+# CI installs libcoraza under /usr/local; a distribution package lands in
+# /usr.  Probe both so the harness is not silently skipped on either.
+my ($coraza_inc) = grep { -f "$_/coraza/coraza.h" }
+    qw(/usr/local/include /usr/include);
+
 plan skip_all => 'coraza headers not available'
-    unless -f '/usr/local/include/coraza/coraza.h';
+    unless defined $coraza_inc;
 
 plan tests => 3;
 
@@ -77,6 +82,18 @@ print {$fh} <<'EOF';
 
 static int force_eof;
 
+/*
+ * Physical end of the backing file.  0 means "the file is as long as the
+ * buffer claims"; a positive value models a response file truncated after
+ * buf->file_last was already recorded, so a read at or past it returns 0.
+ */
+static off_t truncate_at;
+
+/* Bytes handed to coraza_append_response_body() across the whole range. */
+static size_t appended_total;
+static int    append_calls;
+static int    append_fail;
+
 void *
 ngx_pnalloc(ngx_pool_t *pool, size_t size)
 {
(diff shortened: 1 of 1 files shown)


eilandert:
Fixed in `03141f6`. `t/coraza-response-file-eof.t` now covers read errors, inverted ranges, and scratch-buffer allocation failures; allocation, read, and append counters prove each scenario reaches the intended gate. The direct-I/O stubs are limited to `NGX_HAVE_O_DIRECT`, and an `F_NOCACHE` compile control fails when that guard is removed. The focused suites pass 3/3 and 8/8 against nginx 1.31.4, and each of the three production mutations fails its named scenario.

ppomes:
Six of the fixes just landed on `main` (#118, #122–#126), which puts this one in conflict on `t/coraza-response-file-eof.t` (against #122). Could you bring `main` into the branch (a merge, so the history stays as is) when you have a minute? CI and CodeRabbit will re-run on the update and we will merge from there — thanks, and thanks for working through all the review threads.

eilandert:
Merged `main` in (merge, not rebase) — now mergeable at `262934d`.

The conflict in `t/coraza-response-file-eof.t` was not a mechanical one: #122 rewrote the harness around an argument-dispatched `run_filter_case()`, while this branch had extended the previous single-run harness. The two sides cover different production functions, so I resolved it as a union rather than picking a side:

- #122's delayed-response finalize cases (chain-link, stable-buffer, file-read, copy-buffer, copy-data, plus the two success controls) are kept as they are;
- this branch's coverage of `ngx_http_coraza_append_response_body_file()` moved into `run_chunked_reader_cases()`, dispatched as a `chunked-reader` argument, keeping truncation, short reads, mid-file and empty ranges, read errors, inverted ranges and scratch-allocation failure;
- the shared stubs are unioned — `ngx_read_file()`, `ngx_alloc()` and `coraza_append_response_body()` keep this branch's counters and failure injection, `ngx_pnalloc()` keeps #122's `fail_pnalloc` gate;
- the header probe now honours `TEST_LIBCORAZA_INCLUDE` first and otherwise probes `/usr/local/include` then `/usr/include`, so neither the CI layout nor a distribution package is silently skipped.

Plan goes 10 to 11 for the added case.

Verified against nginx 1.31.1 with libcoraza headers under `/usr/include`: 11/11 pass. Removing the premature-EOF gate in the chunked reader fails case 8 and only case 8, so the added coverage is not inert.

ppomes (approved):

